PR #21, opened 2017-05-06
refactoring ODEq and SDEq utils.jl to DEDiffTools

miguelraz:
pew pew!


src/misc_utils.jl
```diff
@@ -8,27 +8,27 @@ Base.@pure function DiffCache{chunk_size}(T, length, ::Type{Val{chunk_size}})
 end
 
 Base.@pure DiffCache(u::AbstractArray) = DiffCache(eltype(u),length(u),Val{ForwardDiff.pickchunksize(length(u))})
```
ChrisRackauckas: Just delete this file

src/StochasticDiffEq.jl
```diff
@@ -20,7 +20,7 @@ module StochasticDiffEq
 
   import DiffEqBase: resize!,deleteat!,addat!,full_cache,user_cache, u_cache,du_cache,
                      resize_non_user_cache!,deleteat_non_user_cache!,addat_non_user_cache!,
-                     terminate!,get_du, get_dt,get_proposed_dt,
+                     terminate!,DiffEqDiffTools.get_du, get_dt,get_proposed_dt,
```
ChrisRackauckas: Don't shadow here. This is a "different `get_du`" which is part of the integrator interface.

src/misc_utils.jl
miguelraz: Deleted

src/StochasticDiffEq.jl
miguelraz: Deleted.

ChrisRackauckas:
You should remove DiffBase, ForwardDiff, and NLsolve from the REQUIRE file.

miguelraz:
Removed.

ChrisRackauckas:
Supersceded by other changes.

PR #26, opened 2017-07-03
Fix and test element-wise tolerances for SRI, SRIW1, SRA, SRA1, and RKMil

devmotion:
Hi!

As suggested in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/71, I applied the same fix to the in-place versions of the above mentioned algorithms.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -28,7 +28,7 @@ function init{uType,tType,isinplace,algType<:Union{AbstractRODEAlgorithm,Abstrac
               qoldinit=1//10^4, fullnormalize=true,
               beta2=beta2_default(alg),
               beta1=beta1_default(alg,beta2),
-              delta=1//6,maxiters = 1e9,
+              delta=map(x->1//6,reltol),maxiters = 1e9,
               dtmax=tType((prob.tspan[end]-prob.tspan[1])),
               dtmin=tType <: AbstractFloat ? tType(10)*eps(tType) : tType(1//10^(10)),
               internalnorm=ODE_DEFAULT_NORM,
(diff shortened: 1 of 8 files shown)


src/solve.jl
```diff
@@ -28,7 +28,7 @@ function init{uType,tType,isinplace,algType<:Union{AbstractRODEAlgorithm,Abstrac
               qoldinit=1//10^4, fullnormalize=true,
               beta2=beta2_default(alg),
               beta1=beta1_default(alg,beta2),
-              delta=1//6,maxiters = 1e9,
+              delta=map(x->1//6,reltol),maxiters = 1e9,
```
ChrisRackauckas: why was reltol introduced here?

devmotion: Because of the type definition of SDEOptions in https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/options_type.jl: `delta` and `reltol` must be of the same type. Without this change `delta` has also to be specified explicitly as a vector if `reltol` is a vector.

But probably a better solution to that problem would be changing the types of `abstol`, `reltol`, and `delta` in https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/options_type.jl to e.g. `A`, `R`, and `D`, respectively.

devmotion:
The tests run fine with the current DiffEqBase master on my computer, so the failures should probably also be fixed with the upcoming DiffEqBase release.

src/solve.jl
ChrisRackauckas: Yes, it would be better to make them different types. I don't think vector delta is sensible.

PR #35, opened 2017-08-27
Do not re-seed RNG in solve()

tkf:
Suggestion: do not re-seed RNG in solve()

I don't think implicitly re-seeding RNG given explicitly by user is the best thing to do.  I think users setting RNG explicitly are likely to prefer controlling the state of RNG.  I suggest to set the seed only when `prob.noise` is `nothing`.

(BTW, the `if prob.seed != 0` block in the old code is suspicious since `if` and `else` block do the same thing.)

One benefit of this patch is that you can do more "rigorous" Monte Carlo simulations by making parallel "independent" streams of RNG using the *jump function*.  The designers of pseudo RNG recommend to use it for parallel simulations (see e.g., http://xoroshiro.di.unimi.it/).  As @ChrisRackauckas mentioned here https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/182#issuecomment-311155957, using the "simulation id" (second argument of `prob_func`) as the seed is probably OK for most use-cases.  But why not allow a better way to do it?  I think this patch is required for this.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -316,14 +316,6 @@ function init(
     if !(typeof(W) <: NoiseGrid) && W.t[end] != t
       error("Starting time in the noise process is not the starting time of the simulation. The noise process should be re-initialized for repeated use")
     end
-    # Reseed
-    if typeof(W) <: NoiseProcess
-      if prob.seed != 0
-        srand(W.rng,seed)
-      else
-        srand(W.rng,seed)
-      end
-    end
   end
 
   EEst = tTypeNoUnits(1)


ChrisRackauckas:
If you get rid of this though, then running the same problem with the same seed twice will give different results.

tkf:
Users can `srand` manually if they want to get the same result. It's more flexible and less magic. With the current implementation, it's not possible to do the right thing like using the jump function.

If you want to keep the magic, why not allow setting `prob.seed = nothing` or something to flag to not re-seed the RNG?

tkf:
BTW, another reason why I was writing this patch is to make the following code work:

```julia
using DifferentialEquations
using RandomNumbers: Xorshifts

# Solve a SDE in one go
noise = WienerProcess(0.0, 0.0, 0.0, rng=Xorshifts.Xoroshiro128Plus(1))
prob = SDEProblem((t, u) -> 0.0, (t, u) -> 1.0, 0.0, (0.0, 2.0), noise=noise)
sol0 = solve(prob)

# Solve the same SDE in two steps
noise = WienerProcess(0.0, 0.0, 0.0, rng=Xorshifts.Xoroshiro128Plus(1))
prob = SDEProblem((t, u) -> 0.0, (t, u) -> 1.0, 0.0, (0.0, 1.0), noise=noise)
sol1 = solve(prob)

prob.tspan = (1.0, 2.0)
prob.u0 = sol1[end]
sol2 = solve(prob)

@assert sol0[end] == sol2[end]
```

That is to say, I thought solving the SDE in one go and solving it in two steps should yield the same result (if the RNG state is carried to the second step).  Is it a bug or it is something to do with integration scheme needing higher order state or something?

Anyway, I think implicit re-seeding is bad since it also makes this impossible.

ChrisRackauckas:
For that example though, shouldn't you be using the integrator interface so that it doesn't reallocate?

>I thought solving the SDE in one go and solving it in two steps should yield the same result (if the RNG state is carried to the second step). Is it a bug or it is something to do with integration scheme needing higher order state or something?

No, that's not what you're doing here. The problem is that you're using an adaptive timestepping algorithm and not taking that into account. One of them steps past 1.0, the other steps directly on it because that's the chosen endpoint. So the same set of random numbers is used differently in these two cases because there's different `dt`s. It's very clear if you plot it:

figure_1-2

The initial `dt` algorithm is conservative and then ramps up. But here you can see that it's running through the random numbers of the other one during its startup phase. 

If you want to match Wiener processes between adaptive runs, you need to use something like a `NoiseWrapper` or a `NoiseGrid` (see the Noise process docs) so that way the values of the Wiener process are aligned and not the random numbers.

>If you want to keep the magic, why not allow setting prob.seed = nothing or something to flag to not re-seed the RNG?

That sounds good. My thing is that requiring the user to create a noise process just to run the same seed seems like overkill, but I do want to give the user full control. This seems like a good compromise.

tkf:
Ah, I see.  Thank you for the clarification of what is happening!  I plotted the solution and realized the little wiggles in `sol2` but couldn't connect it to adaptive time step.

Regarding the API, another option I can think of is to let `NoiseProcess` accept the RNG instance *and* RNG factory; i.e., something like this:

```julia
noise0 = WienerProcess(..., rng_factory=Xorshifts.Xoroshiro128Plus)
prob0 = SDEProblem(..., noise=noise0, seed=1)
solve(prob0)  # seed the RNG

noise1 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus(1))
prob1 = SDEProblem(..., noise=noise1)
solve(prob1)  # don't seed the RNG
```

Or you can even let `rng=` accept both the constructor and instance and seed it only when the constructor is given.  I think accepting the constructor is closer to your intention since current released code only use the information of the type, not the state.  I actually like this API since it lets users know that the state of the RNG is set somewhere else when the constructor is given.  But the problem is this API breaks users' code.

I don't know which API is more consistent with other parts of DifferentialEquations.jl.  Which API is better?

ChrisRackauckas:
>Or you can even let rng= accept both the constructor and instance and seed it only when the constructor is given. I think accepting the constructor is closer to your intention since current released code only use the information of the type, not the state.

In the first case I guess we just give it a default seed and then check for this and reseed in the integrator? While in the second we see if it was given with a seed and then don't touch it? I think the noise process will need a bool to keep track of this, but I think it'll work.

tkf:
To be on the same page, let's write down what would happen in those APIs with various combinations of arguments.

## Candidate 1 (`seed :: Union{Int, Void}`)

```julia
noise00 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus(1))
prob00 = SDEProblem(..., noise=noise00)
solve(prob00)  # seed the RNG in a non-reproducible way

noise01 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus(1))
prob01 = SDEProblem(..., noise=noise01, seed=2)
solve(prob01)  # re-seed the RNG (=> seed=2 is used)

noise10 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus(1))
prob10 = SDEProblem(..., noise=noise10, seed=nothing)
solve(prob1)  # don't seed the RNG (=> seed=1 is used)
```

(`prob11` in Candidate 2 (below) is identical to `prob01` in Candidate 1)

### Implementation

```julia
if prob.seed != nothing
  srand(rng, (prob.seed == 0 ? rand(UInt64) : prob.seed))
end
```

## Candidate 2 (`rng :: Union{AbstractRNG, DataType, Function}`)

```julia
noise00 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus)
prob00 = SDEProblem(..., noise=noise00)
solve(prob00)  # create a RNG in a non-reproducible way

noise01 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus)
prob01 = SDEProblem(..., noise=noise01, seed=2)
solve(prob01)  # create a RNG with given seed (=> seed=2 is used)

noise10 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus(1))
prob10 = SDEProblem(..., noise=noise10)
solve(prob10)  # don't seed the RNG (=> seed=1 is used)

noise11 = WienerProcess(..., rng=Xorshifts.Xoroshiro128Plus(1))
prob11 = SDEProblem(..., noise=noise11, seed=2)
solve(prob11)  # re-seed the RNG (=> seed=2 is used)
```

### Implementation

```julia
if isa(rng, AbstractRNG)
  rng = rng(prob.seed == 0 ? rand(UInt64) : prob.seed)
end
```

or even

```julia
if isa(rng, AbstractRNG)
  rng = prob.seed == 0 ? rng() : rng(prob.seed)
end
```

## Candidate 2' (`rng :: AbstractRNG`, `rng_factory :: Union{DataType, Function}`)

Just like Candidate 2, but `rng` accepts only `AbstractRNG`.  To pass RGN factory, use `rng_factory`.

## Notes

Above type assertions are just for illustration and not super exact; e.g., I think `seed` should also accept a string so that we can do `seed="/dev/random"`.

tkf:
I guess we should fix `solve(::AbstractNoiseProblem)` at the same time. It does `deepcopy` noise process, including RNG (https://github.com/JuliaDiffEq/StochasticDiffEq.jl/issues/38#issuecomment-325261386) but `solve(::SDEProblem)` doesn't. If we go with RNG factory, I don't think we need deepcopy (unless noise process holds some other states).

ChrisRackauckas:
Reference: https://discourse.julialang.org/t/how-do-i-deal-with-random-number-generation-when-multithreading/5636

I hope to get to this soon.

tkf:
Thanks.  Just FYI, I don't think `randjump` is implemented in RandomNumbers.jl yet.  But I think the author is working on it since "Take parallel into account." is in the list of millstones: https://github.com/sunoru/RandomNumbers.jl/issues/3

If you want `MonteCarloProblem` to support it, you may need to go back to Mersenne Twister at the moment.

ChrisRackauckas:
Fixed by https://github.com/JuliaDiffEq/StochasticDiffEq.jl/commit/c4e70552466bb44e973f4a6681ed8f04e819e8e4 . I just made it an option that you can turn off. It defaults to on so that if you have everything to reset and gave it a seed it will automatically cause it to solve the same process. On the otherhand, you can tell it to not reset, and even if you reset it (to clear the previous memory of the noise process) you can tell it to not reseed. So then with that off you can do the kind of jump seeding mentioned in those threads.

PR #53, opened 2018-02-14
Predictor Corrector method 

onoderat:
The predictor corrector method for SDEs have been implemented and tested. 

Both the in place and not in place versions have been implemented. 
The testing code used the problem from the commutative_tests.jl . 

The speed of the code has not been benchmarked rigorously. For a single setting of dt=0.002, the following speeds were achieved. 
```
    EM in place            : 1.665ms (21771 allocations: 2.01 MiB)
    RKMil in place         : 1.881ms (23776 allocations: 2.10 MiB)
    PCEuler not in place   : 2.626ms (40673 allocations: 3.44 MiB)
    PCEuler in place       : 1.942ms (24271 allocations: 2.16 MiB)
```
The convergence tests shows that both the PC has better accuracy than EM and surprisingly also RKMilCommute. Although the PC is proven to be order 0.5, it showed order 1 scaling for this particular problem (see figure below). At dt=1/2^10, the l2 errors are respectively
    RKMil        : 6.02e-3
    PCEuler    : 1.44e-3

error_scaling


src/algorithms.jl
```diff
@@ -22,6 +22,35 @@ Base.@pure RKMil(;interpretation=:Ito) = RKMil{interpretation}()
 struct RKMilCommute{interpretation} <: StochasticDiffEqAlgorithm end
 Base.@pure RKMilCommute(;interpretation=:Ito) = RKMilCommute{interpretation}()
 
+###############################################################################
+
+# Predictor Corrector
+struct PCEuler{T<:Real} <: StochasticDiffEqAlgorithm
+  theta::T
+  eta::T
+  bbprime::Function
```
ChrisRackauckas: Parameterize this for stability purposes.

src/algorithms.jl
```diff
@@ -22,6 +22,35 @@ Base.@pure RKMil(;interpretation=:Ito) = RKMil{interpretation}()
...
+        If problem is in place - a in place bbprime should be supplied - and vice versa for not in place speicification of problem.
+kwargs: theta::Real
+          Degree of implicitness in the drift term. Set to 0.5 by default.
+        eta::Real
+          Degree of implicitness in the diffusion term. Set to 0.5 by default.
+
+Reference: Stochastics and Dynamics, Vol. 8, No. 3 (2008) 561–581
+Note that the original paper has a typo in the definition of bbprime...
+"""
+# TODO: Check if this can be implemented with @pure.
```
ChrisRackauckas: This doesn't need `@pure`. It'll infer without it.

src/perform_step/predcorr.jl
```diff
@@ -0,0 +1,70 @@
...
+  @unpack theta,eta,bbprime = integrator.alg
+  dW = W.dW
+  if !is_diagonal_noise(integrator.sol.prob) || typeof(W.dW) <: Number
+    gdW_n = g(uprev,p,t)*dW
+  else
+    gdW_n = g(uprev,p,t).*dW
+  end
+
+  f_n = f(uprev,p,t)
+  ubar = @muladd uprev + dt*f_n + gdW_n
```
ChrisRackauckas: `@muladd` on the function makes everything inside `@muladd`, so these aren't needed (there are probably some other functions in this library that need cleaning for that)

src/perform_step/predcorr.jl
```diff
@@ -0,0 +1,70 @@
...
+  tnp1 = t + dt
+  fbar_np1 = @muladd f(ubar, p, tnp1) - eta*bbprime(ubar, p, tnp1)
+
+  if !is_diagonal_noise(integrator.sol.prob) || typeof(dW) <: Number
+    gdW_np1 = g(ubar,p,tnp1)*dW
+  else
+    gdW_np1 = g(ubar,p,tnp1).*dW
+  end
+
+  u .= @muladd uprev + (theta*dt)*fbar_np1 + ((1-theta)*dt)*fbar_n + eta*gdW_np1 + (1-eta)*gdW_n
```
ChrisRackauckas: You don't want to `.=` for numbers and static arrays. It'll work but be slower.

test/predcorr_tests.jl
```diff
@@ -0,0 +1,89 @@
+using StochasticDiffEq, DiffEqProblemLibrary, DiffEqDevTools, Base.Test
```
ChrisRackauckas: Just rename this file as `nondiagonal_tests.jl`: I assume other algorithms will get added to it later.

ChrisRackauckas:
Looks great! Just a few minor changes and I'll merge. I'm happily surprised it does that much better on this problem: that's a very big improvement! I wonder if it's order 1 on commutative noise problems or something like that: that could make sense.

src/algorithms.jl
onoderat: Could you elaborate on this a little more? Would you like the PCEuler to be parametrized by the  type of the function bbprime?

src/algorithms.jl
onoderat: Thanks - I will remove the comment.

src/perform_step/predcorr.jl
onoderat: Awesome - I did note that adding this did not change the speed. Will remove them.

src/perform_step/predcorr.jl
onoderat: Sure, I will copy what you did for EM then.

test/predcorr_tests.jl
onoderat: Will do.

src/algorithms.jl
ChrisRackauckas: In Julia, every function is a different type and `Function` is just an abstract type. This allows specialization on functions and then allows things like inlining. If you `::Function` in a type, then it just knows it's a `<:Function` and cannot optimize directly on it. So yup, just make that `::F` and a free-roaming `F` parameter.

src/algorithms.jl
```diff
@@ -22,6 +22,35 @@ Base.@pure RKMil(;interpretation=:Ito) = RKMil{interpretation}()
...
+PCEuler(bbprime::Function; theta::T=1/2, eta::T=1/2) where {T<:Real} =
```
ChrisRackauckas: Don't `::Function` in a type signature. It has a non-intuitive anti-specialization behavior, but also not all functions `<:Function`.

onoderat: Wow, I have been doing that for a while now (in my other code) thanks for the tip! Will change.

onoderat: Ah that reminds me that I saw you did something like that to define the analytic solution to f. Ok will remove the ::Function in the type signature.

src/algorithms.jl
onoderat: I see, will make the change.

ChrisRackauckas:
Just noticed `bbprime` should change to `ggprime` since the rest of the docs and functions refer to the diffusion term by `g`.

ChrisRackauckas:
Good stuff, thanks. After seeing that from RKMil, I wonder how much of that is from the RK handling of the derivative handling. Would you mind doing a PR to do `ggprime` implementations of Milstein?

ChrisRackauckas:
This needs to get added to the docs page http://docs.juliadiffeq.org/latest/solvers/sde_solve.html . It should probably be recommended when `ggprime` is known, and the definition of `ggprime` will need to go somewhere there for now. Eventually we'll want to handle `ggprime` like in:

http://docs.juliadiffeq.org/latest/features/performance_overloads.html

but that should wait on the change to `DiffEqFunction` where we can then have specific setups that are different for the different problem types.

onoderat:
Hi Chris - I would like to do the ggprime implementation of Milstein. The issue is that the one in the textbook does not have implicit noise and neither do any papers explicitly construct this. 

I am a physicist so I cant really do the proofs for these methods. Perhaps, you want to have a stab at the proof? Once I have the formula, I can do a PR for the method.

ChrisRackauckas:
What do you mean by implicit noise? This shouldn't have any implicit terms.

onoderat:
Ah sorry - I meant implicitness in the diffusion. Basically the one in the textbook does not have a eta parameter - it only has implicitness in the drift. See last sentence of screenshot below -

screen shot 2018-02-14 at 11 07 32 am

ChrisRackauckas:
I was thinking of doing the simpler one first:

milstein

onoderat:
Ah I see. Sure that should be easy to code up. But currently I am kind of rushing to prepare for the APS march meeting. Perhaps we could discuss this further when we meet in person next month? 

I am thinking of working on the Wiktorsson approximations and creating a better interface for that stuff through NoiseProcess. Once that is done, perhaps we can multiple dispatch on the type of problem (commutative, diagonal, etc...). 

This way we do not need to define RKMIlCommute, RKMil etc. What do you think?

ChrisRackauckas:
>I am thinking of working on the Wiktorsson approximations and creating a better interface for that stuff through NoiseProcess. Once that is done, perhaps we can multiple dispatch on the type of problem (commutative, diagonal, etc...). This way we do not need to define RKMIlCommute, RKMil etc. What do you think?

Yeah, share your idea when you can. As always, there's no rush. Could you open an issue to discuss this? I'm very curious about your idea here. I want to do something for this, and we should have a more constrained version of jumps that allow for these kinds of schemes from Platen.

ChrisRackauckas:
BTW, schemes are usually drift-implicit because the inverse of the normal has infinite moments making it hard to prove convergence. There's things like truncated schemes though to get around this.

onoderat:
Sure I will open an issue on this. 

Thanks for explaining why the diffusion-implicit is hard. Interesting for the predictor-corrector, I numerically found that eta=0.5 is critical for getting that order 1 scaling. 

Perhaps its an interesting research direction to investigate whether the same higher order scaling can be done if we have an eta for the milstein case? Like say we get order 1.5 for doing order 1 level work? 

screen shot 2018-02-14 at 12 06 57 pm

ChrisRackauckas:
That's super convergence. A lot of it is just finding the right conditions to make convergence "better", like how on additive noise the Euler-Maruyama scheme is Order 1.0. It would be very interesting to find out super convergence criterias for high order SDE schemes. I would assume that many of the common categories (additive, multiplicative) have some super convergence stuff going on.

PR #56, opened 2018-03-27
Implement set_{u,t,ut}! for SDEIntegrator

tkf:
Add the interface defined by https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/86

Mostly copy & paste of https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/256 at the moment. I guess there should be some tests.

### DIFF
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -332,3 +332,33 @@ end
 @inline function DiffEqBase.get_du!(out,integrator::SDEIntegrator)
   @. out = (integrator.u - integrator.uprev) / integrator.dt
 end
+
+function DiffEqBase.set_t!(integrator::SDEIntegrator, t::Real)
+  if integrator.opts.save_everystep
+    error("Integrator time cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
+  if !isdtchangeable(integrator.alg)
+    reinit!(integrator, integrator.u;
+            t0 = t,
+            reset_dt = false,
+            reinit_callbacks = false,
+            reinit_cache = false)
+  else
+    integrator.t = t
+  end
+end
+
+function DiffEqBase.set_u!(integrator::SDEIntegrator, u)
+  if integrator.opts.save_everystep
+    error("Integrator state cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
+  integrator.u = u
+  u_modified!(integrator, true)
+end
+
+function DiffEqBase.set_ut!(integrator::SDEIntegrator, u, t::Real)
+  DiffEqBase.set_u!(integrator, u)
+  DiffEqBase.set_t!(integrator, t)
+end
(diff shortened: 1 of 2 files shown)


src/integrators/integrator_interface.jl
```diff
@@ -332,3 +332,33 @@ end
...
+    error("Integrator state cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
+  integrator.u = u
+  u_modified!(integrator, true)
+end
+
+function DiffEqBase.set_ut!(integrator::SDEIntegrator, u, t::Real)
+  DiffEqBase.set_u!(integrator, u)
+  DiffEqBase.set_t!(integrator, t)
```
tkf: This is exactly same as the one in the ODE.  Maybe this should be implemented in DiffEqBase as the default?

src/integrators/integrator_interface.jl
```diff
@@ -332,3 +332,33 @@ end
 @inline function DiffEqBase.get_du!(out,integrator::SDEIntegrator)
   @. out = (integrator.u - integrator.uprev) / integrator.dt
 end
+
+function DiffEqBase.set_t!(integrator::SDEIntegrator, t::Real)
+  if integrator.opts.save_everystep
+    error("Integrator time cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
+  if !isdtchangeable(integrator.alg)
```
tkf: There was `alg_extrapolates(integrator.alg) ||` in the ODE version. Anything like that in SDE?

ChrisRackauckas: Nope, there's no extrapolating in the SDEs.

tkf: No other traits that prevent non-`reinit!` version of `set_t!`?

ChrisRackauckas: I don't think there is. Comment out the `isdtchangable` though since that could show up if a multistep SDE method is implemented, which I do hope to get to.

ChrisRackauckas: The algorithm traits in OrdinaryDiffEq.jl were added as needed, but now that we kind of have a good set it may be useful to up that to some more general DiffEq thing.

tkf: > Comment out the isdtchangable

What do you mean?  Just use `integrator.t = t` only?

I see that `isdtchangable` always returns true (except the methods in BridgeDiffEq.jl?).  So, maybe you mean you don't want a "dead path" in the function?  Comment out the `reinit!` path until you implement other methods?

ChrisRackauckas: oh, I didn't realize `isdtchangable` for SDEs already exists haha. I guess I do have it there just for the possible multistep methods in the future, so it's fine.

ChrisRackauckas:
I'd say just move `set_ut!` to DiffEqBase and it's good.

src/integrators/integrator_interface.jl
tkf: I see!

ChrisRackauckas:
Thanks! Merging since I know it will pass after 3.8.0 merge.

tkf:
Thanks!

PR #67, opened 2018-04-01
Noise implicit methods

ChrisRackauckas:


ChrisRackauckas:
@Gaussia you will probably want to try the `ISSEM` algorithm from this when it merges.

PR #72, opened 2018-04-13
Use DiffEqBase.check_error!

tkf:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/276


ChrisRackauckas:
Should restart after https://github.com/JuliaLang/METADATA.jl/pull/14250

PR #74, opened 2018-04-14
Call @sde_exit_conditions in step!

tkf:
See: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/318


ChrisRackauckas:
Same fix here?

tkf:
Tests on Travis passed https://travis-ci.org/JuliaDiffEq/StochasticDiffEq.jl/jobs/366378146#L1166 but hit the time limit during coverage analysis.

PR #77, opened 2018-04-15
Add eigenvalue estimation for `SOSRI2` and `SOSRA2`

YingboMa:


src/perform_step/sra.jl
```diff
@@ -189,6 +189,12 @@ end
 
   u = uprev + E₁ + E₂ + W.dW*(beta11*g1 + beta12*g2 + beta13*g3)
 
+  if typeof(integrator.alg) <: StochasticCompositeAlgorithm && integrator.alg.algs[1] === SOSRA2()
```
ChrisRackauckas: `typeof(integrator.alg.algs[1]) <: SOSRA2()`

ChrisRackauckas: That avoids having to build the type, which may not be elided. I think it might be now, but if we add options...

ChrisRackauckas:
MWTP

PR #78, opened 2018-04-15
Add stiffness detection and automatic switching algorithm

YingboMa:


src/alg_utils.jl
```diff
@@ -34,7 +34,9 @@ alg_order(alg::SRA3) = 2//1
 alg_order(alg::SOSRA) = 2//1
 alg_order(alg::SOSRA2) = 2//1
 alg_order(alg::SKenCarp) = 2//1
-alg_order(alg::Union{StochasticDiffEqCompositeAlgorithm,StochasticDiffEqRODECompositeAlgorithm}) = alg_order(alg.algs[1])
+alg_order(alg::Union{StochasticDiffEqCompositeAlgorithm,StochasticDiffEqRODECompositeAlgorithm}) = alg_order(alg.algs[alg.current_alg])
```
ChrisRackauckas: shouldn't use current_alg

src/algorithms.jl
```diff
@@ -265,9 +265,11 @@ Base.@pure SKenCarp(;chunk_size=0,autodiff=true,diff_type=Val{:central},
 
 # Etc.
 
-struct StochasticCompositeAlgorithm{T,F} <: StochasticDiffEqCompositeAlgorithm
+mutable struct StochasticCompositeAlgorithm{T,F} <: StochasticDiffEqCompositeAlgorithm
   algs::T
   choice_function::F
+  current_alg::Int
```
ChrisRackauckas: this shouldn't be necessary anywhere?

PR #80, opened 2018-05-14
Refactor implicit solvers (1)

YingboMa:


YingboMa:
I am not sure about a way to implement a common nonlinear solver. The nonlinear solver seems to be more flexible in `StochasticDiffEq.jl`.

ChrisRackauckas:
What do you mean?

ChrisRackauckas:
MWTP

PR #88, opened 2018-06-18
Fix some typos in `PCEuler` docs

darwindarak:
Preempt a `syntax: invalid escape sequence` error in Julia v0.7


ChrisRackauckas:
Cool thanks! Another thing that can be done is adding `doc` before the `"""` since the `@doc_str` macro I think fixes it.

PR #89, opened 2018-07-12
Change save_start to take saveat into consideration

Vaibhavdixit02:


ChrisRackauckas:
`FunctionMap` is only an algorithm in OrdinaryDiffEq.jl.

PR #90, opened 2018-07-14
0.7 compatibility

YingboMa:


test/adaptive/sde_lorenzattractor_tests.jl
```diff
@@ -1,4 +1,5 @@
-using StochasticDiffEq, DiffEqProblemLibrary
+using StochasticDiffEq, Random
+import DiffEqProblemLibrary.SDEProblemLibrary: prob_sde_lorenz
```
ChrisRackauckas: doesn't that importsdeproblems() need to be ran?

PR #93, opened 2018-07-24
Mass matrix hotfix

MSeeker1340:
Address changes in JuliaDiffEq/DiffEqBase.jl#126.

I think the `SDEProblem` and `SDEFunction` interface can still use some improvements though. It doesn't really make sense to repeat the `g` field in both the problem and the function --- one should be enough.


ChrisRackauckas:
>I think the SDEProblem and SDEFunction interface can still use some improvements though. It doesn't really make sense to repeat the g field in both the problem and the function --- one should be enough.

Indeed. I just haven't worked through the whole thing yet.

ChrisRackauckas:
Looks like there was some issues with the local testing setup so I got this from here. Thanks!

PR #99, opened 2018-08-18
Upgrade implicit solvers to the latest nlsolve

YingboMa:


ChrisRackauckas:
Looks like this is very far along, just some test setup issues now?

https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/a575b2211421dce96be59031bb5ea8a44bb51b08/test/sde/sde_convergence_tests.jl#L67-L68

ISSEM is a 1/2 order algorithm. ImplicitEM is a 1/2 order algorithm. ImplicitRKMil is an order 1 algorithm. So the tests 

https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/a575b2211421dce96be59031bb5ea8a44bb51b08/test/sde/sde_convergence_tests.jl#L67-L72

have too high of convergence coefficients, which seems to be the failure.

ChrisRackauckas:
Actually on additive noise those should all be 1.0.

src/perform_step/sdirk.jl
```diff
@@ -5,21 +5,20 @@
...
   alg = unwrap_alg(integrator, true)
   theta = alg.theta
   alg.symplectic ? a = dt/2 : a = dt
-  uf.t = t
   γdt = dt*theta
   repeat_step = false
-  J, W = calc_W!(integrator, cache, γdt, repeat_step)
+  if nlsolve! isa NLNewton
+    J, nlcache.W = calc_W!(integrator, cache, γdt, repeat_step)
+    uf.t = t
```
ChrisRackauckas: this should go before?

src/perform_step/kencarp.jl
```diff
@@ -19,125 +17,61 @@
...
-
   γdt = γ*dt
 
   # calculate W
-  uf.t = t
   repeat_step = false
-  J, W = calc_W!(integrator, cache, γdt, repeat_step)
+  if nlsolve! isa NLNewton
+    J, nlcache.W = calc_W!(integrator, cache, γdt, repeat_step)
+    uf.t = t
```
ChrisRackauckas: this should go before?

ChrisRackauckas:
Upping the number of Monte Carlo runs got rid of the superconvergence and made the results match the theoretical results perfectly, so I think those are good now. Fingers crossed tests pass.

ChrisRackauckas:
Something seems wrong in the split tests

PR #105, opened 2018-09-27
Remove SpecialMatrices & callbacks saving fix

MSeeker1340:
The SpecialMatrices package is no longer available due to the recent package compatibility capping and it is preventing tests from executing. The only affected test script is multivariate_geometric.jl which for some reason is not included in the test suite. I have added it and also made some minor fixes to the test script (note that the last test seems to be broken).

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -236,8 +236,11 @@ end
 function apply_discrete_callback!(integrator::SDEIntegrator,callback::DiscreteCallback)
   saved_in_cb = false
   if callback.condition(integrator.u,integrator.t,integrator)
+    # handle saveat
+    savevalues!(integrator)
     if callback.save_positions[1]
-      savevalues!(integrator,true)
+      # if already saved then skip saving
+      last(integrator.sol.t) == integrator.t || savevalues!(integrator,true)
       saved_in_cb = true
     end
     integrator.u_modified = true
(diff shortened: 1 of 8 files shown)


MSeeker1340:
Since the fix for saveat handling in callbacks is small let's just merge these together.

MSeeker1340:
For some reason the first test (which passes on my local machine) fails on CI, so I'm just gonna flag it anyways. Someone should take a look at the multivariate geometric tests in the future.

Also updated the CI scripts to use Julia v1.0.

ChrisRackauckas:
Multivariate geometric is using the average over 50 runs to calculate convergence. Try 100 or 200.

MSeeker1340:
I added `Random.seed!` to ensure the first test passes. However the second test is still broken even after increasing to 100 runs (200 runs takes too long to run).

src/callbacks.jl
```diff
@@ -196,8 +196,12 @@ function apply_callback!(integrator,callback::ContinuousCallback,cb_time,prev_si
     change_t_via_interpolation!(integrator,integrator.tprev+cb_time)
   end
 
+  # handle saveat
+  savevalues!(integrator)
```
ChrisRackauckas: Just for continuous callbacks?

MSeeker1340: Ah yes I forgot to include Yingbo's fix on discrete callbacks. Let me fix that.

ChrisRackauckas:
It looks like the CI update is pointing out that there are some floating point issues coming up, but it's unrelated to this PR so I'll merge and handle it later.

PR #116, opened 2018-11-05
Specialize get_tmp_cache

devmotion:
This fixes https://github.com/JuliaDiffEq/StochasticDiffEq.jl/issues/115 by more or less copying the approach in OrdinaryDiffEq. Still needs some tests.


ChrisRackauckas:
We should probably get a test on this too, but it seems fine.

PR #119, opened 2018-11-24
mirror performance fix by parameterizing nlsolve caches

ChrisRackauckas:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/565 @yingboma

### DIFF
--- a/src/caches/sdirk_caches.jl
+++ b/src/caches/sdirk_caches.jl
@@ -42,7 +42,6 @@ DiffEqBase.@def iipnlcachefields begin
     reltol = 1e-1 # TODO: generalize
     tol = uToltype(min(0.03,first(reltol)^(0.5)))
   end
-  _nlsolve = alg.nlsolve
 end
 
 DiffEqBase.@def oopnlcachefields begin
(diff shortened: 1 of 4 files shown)


src/nlsolve/type.jl
```diff
@@ -20,25 +20,34 @@ mutable struct NLSolverCache{rateType,uType,W,uToltype,cType,gType} <: AbstractN
...
+function NLAnderson(n=5; kwargs...)
+  nlcache = NLSolverCache(;kwargs...)
+  NLAnderson{true, typeof(nlcache)}(nlcache, n)
+end
+function NLNewton(;kwargs...)
+  nlcache = NLSolverCache(;kwargs...)
+  NLNewton{true, typeof(nlcache)}(nlcache)
+end
 
 oop_nlsolver(s::NLFunctional{true}) = NLFunctional{false}(s.cache)
```
YingboMa: These two methods are no longer needed.

src/caches/sdirk_caches.jl
```diff
@@ -75,6 +75,17 @@ DiffEqBase.@def oopnlcachefields begin
   _nlsolve = oop_nlsolver(alg.nlsolve)
```
YingboMa: This line is no longer needed together with https://github.com/JuliaDiffEq/StochasticDiffEq.jl/pull/119/files#diff-764be46c2a94bdca1a4b0bc276f043d4R45.

PR #120, opened 2018-12-14
Use `internalnorm` instead of `abs` in lamba.jl

Krastanov:
Permits use of the Lamba integrators with complex arrays.

Partially addresses https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/401

### DIFF
--- a/src/perform_step/lamba.jl
+++ b/src/perform_step/lamba.jl
@@ -76,7 +76,7 @@ end
 
 
     @tight_loop_macros for (i,atol,rtol) in zip(eachindex(u),Iterators.cycle(integrator.opts.abstol),Iterators.cycle(integrator.opts.reltol))
-      @inbounds tmp[i] = (tmp[i])/(atol + max(abs(uprev[i]),abs(u[i]))*rtol)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Yeah, OrdinaryDiffEq switched everything from abs to internalnorm for the complex handling, but it never got to all of the SDE solvers.

Krastanov:
I looked up the AppVeyor failure report, but it does not mean much to me.

As for the other solvers (at least the non-stiff ones), I did not get any other `abs` errors for them (see https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/401), however I did get other errors.

devmotion:
> As for the other solvers (at least the non-stiff ones), I did not get any other abs errors for them (see JuliaDiffEq/DifferentialEquations.jl#401), however I did get other errors.

Since the residuals are computed at the end of each step, most likely other failures mask the `internalnorm` errors.

ChrisRackauckas:
Thanks! I went ahead and applied these changes to the other methods.

PR #130, opened 2019-01-23
Clean and update solve.jl

devmotion:
When I started debugging the stack overflow error that I get when I replace `diffeq_fd` with `diffeq_rd` in the [DiffEqFlux example](https://github.com/JuliaDiffEq/DiffEqFlux.jl#using-other-differential-equations) (by the way, e.g., using `EM` (with fixed step size `dt = 0.01`) and `SRIW1` with `diffeq_rd` works fine at least in this example), I noticed that solve.jl contains some redundant and unneeded lines and differs from the implementation in OrdinaryDiffEq quite a bit although it shouldn't (it seems at least) - I assume it's just different updates and improvements not being applied to StochasticDiffEq as well. I tried to fix these issues.

I was wondering whether parts of the implementation in OrdinaryDiffEq and StochasticDiffEq could/should be moved to a common package (maybe DiffEqBase but I guess implementations tailored only for OrdinaryDiffEq and StochasticDiffEq might be a bit too specific there?) to avoid such issues in the future; e.g., `handle_dt!` is completely identical, large parts of the argument parsing and checking, type construction are identical or at least similar (by moving specific conditions that are different in both packages to separate functions one could increase this even more, I guess), and I assume that also other parts of the integrator interface are (or should be) implemented in a identical or very similar way.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -131,8 +131,6 @@ function DiffEqBase.__init(
   dtmax > zero(dtmax) && tdir < 0 && (dtmax *= tdir) # Allow positive dtmax, but auto-convert
   # dtmin is all abs => does not care about sign already.
 
-  order = typeof(alg) <: SRI || typeof(alg) <: SRA ? alg.tableau.order : alg_order(alg)
-
   if isinplace(prob) && typeof(u) <: AbstractArray && eltype(u) <: Number && uBottomEltypeNoUnits == uBottomEltype # Could this be more efficient for other arrays?
     if !(typeof(u) <: ArrayPartition)
       rate_prototype = similar(u,typeof(oneunit(uBottomEltype)/oneunit(tType)))
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -1,69 +1,74 @@
-function __solve(
-  prob::AbstractRODEProblem,
-alg::algType,timeseries=[],ts=[],ks=[],recompile::Type{Val{recompile_flag}}=Val{true};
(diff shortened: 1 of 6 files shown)


src/solve.jl
```diff
@@ -1,98 +1,112 @@
...
   beta1=beta1_default(alg,beta2),
-  delta=1//6,maxiters = 1e9,
-  dtmax=eltype(tType)((prob.tspan[end]-prob.tspan[1])),
-  dtmin=eltype(tType) <: AbstractFloat ? eps(eltype(tType)) : eltype(tType)(1//10^(10)),
-  internalnorm=ODE_DEFAULT_NORM,
-  unstable_check = ODE_DEFAULT_UNSTABLE_CHECK,
+  delta=1//6,
+  maxiters = 1000000,
+  dtmax=eltype(prob.tspan)((prob.tspan[end]-prob.tspan[1])),
+  dtmin = typeof(one(eltype(prob.tspan))) <: AbstractFloat ? 10*eps(eltype(prob.tspan)) :
```
ChrisRackauckas: eps seems to work better

devmotion: Oh I've missed this change in OrdinaryDiffEq

ChrisRackauckas: yeah, the reason is 10eps(Float32) > 1e-6 which is the minimum starting dt the auto-dt algorithm gives, so you get instant "below mindt" errors on some problems if you're using Float32s. While dropping the 10 doesn't seem to actually effect Float64 code, since if you're at dt=1e-15 you tend to go unstable to dt->0 anyways.

PR #132, opened 2019-01-28
Add `vec` and `reshape` in out-of-place stiff solvers

YingboMa:


devmotion:
Complex tests fail due to a recent update of DiffEqNoiseProcess: https://github.com/JuliaDiffEq/DiffEqNoiseProcess.jl/pull/15

PR #134, opened 2019-02-09
Fix https://github.com/JuliaDiffEq/DiffEqBase.jl/issues/188

ChrisRackauckas:


isaacsas:
I think this failed as this was from back when there were Meta -> General Registry issues going on. Should it work and be mergeable now?

Issue #65, opened 2018-04-01
Split Step Explicit Methods


ChrisRackauckas:
These are a lot like the methods that @onoderat implemented, but they are fully explicit and don't require derivative calculations. They are essentially the Runge-Kutta forms of that. They aren't too difficult to implement (but could be extremely impactful), so for now I'll leave the Milstein methods as a possible outside contribution for others.

https://www.sciencedirect.com/science/article/pii/S0377042709007419

ChrisRackauckas:
The Euler methods are done via just putting a splitting option in EM and LambaEM

PR #150, opened 2019-05-29
Tried implementing split-step explicit method

deeepeshthakur:
@ChrisRackauckas I tried methods from #65 
Implemented the TSM1a. In the paper it's shown that it has an order of 1.0 but I'm getting 0.5. I can't find what I did wrong.
[converge_tests](https://htmlpreview.github.io/?https://github.com/deeepeshthakur/Urls/blob/master/WangLiTSM1A.html)
[paper](https://reader.elsevier.com/reader/sd/pii/S0377042709007419?token=633B23BFA790AEC9EA0584079AAAE66632530C0E05C66B980B94D4A71F327644F8069723D6871EC235B7BA4CEBCC898A)

And right now I've not added any tests


src/perform_step/low_order.jl
```diff
@@ -230,3 +230,50 @@ end
...
+
+#methods from https://doi.org/10.1016/j.cam.2009.11.010
+@muladd function perform_step!(integrator,cache::WangLi3SMil_AConstantCache,f=integrator.f)
+  @unpack t,dt,uprev,u,W,p = integrator
+  #stage 1
+  k  = integrator.f(uprev,p,t)
+  u  = uprev + dt*k
+
+  #stage 2
+  k  = integrator.g(u,p,t)
```
ChrisRackauckas: dividing by sqdt means that the effective diffusion distance is like dt, which means you want this evaluated at t, and so this should be `g(uprev,p,t)`

src/perform_step/low_order.jl
```diff
@@ -230,3 +230,50 @@ end
...
+  tmp = u + k*integrator.sqdt
+  k₁ = integrator.g(tmp,p,t)
+  k₁ = (k₁ - k)/(integrator.sqdt)
+  u  = u - (dt/2)*k*k₁
+
+  #stage 3
+  k  = integrator.g(u,p,t)
```
ChrisRackauckas: `g(uprev,p,t)` here too. Cache it so it's only done once.

src/perform_step/low_order.jl
deeepeshthakur: image
Basically this is the method.I don't want it evaluated at uprev.

ChrisRackauckas: You're not exactly implementing that method though. You're doing a finite difference approximation to the g'. That's fine, but what I'm mentioning is that you need `k  = integrator.g(uprev,p,t)` in order to do that finite difference approximation between two `g` calculations that are `dt` spaced apart.

src/perform_step/low_order.jl
```diff
@@ -230,3 +230,50 @@ end
...
+@muladd function perform_step!(integrator,cache::WangLi3SMil_AConstantCache,f=integrator.f)
+  @unpack t,dt,uprev,u,W,p = integrator
+  #stage 1
+  k  = integrator.f(uprev,p,t)
+  u  = uprev + dt*k
+
+  #stage 2
+  k  = integrator.g(u,p,t)
+  tmp = u + k*integrator.sqdt
+  k₁ = integrator.g(tmp,p,t)
```
ChrisRackauckas: t+dt

src/perform_step/low_order.jl
```diff
@@ -230,3 +230,50 @@ end
...
+  k₁ = (k₁ - k)/(integrator.sqdt)
+  u  = u - (dt/2)*k*k₁
+
+  #stage 3
+  k  = integrator.g(u,p,t)
+  tmp = u + k*integrator.sqdt
+  k₁ = integrator.g(tmp,p,t)
```
ChrisRackauckas: t+dt

deeepeshthakur:
@ChrisRackauckas I tried another method from the same paper but it has the same convergence of 0.5

src/perform_step/low_order.jl
deeepeshthakur: @ChrisRackauckas This is from a previous paper by the same author describing 2-stage Milstein methods
image
There they assume g = g(t,u) and they use g(t,Yn)

src/perform_step/low_order.jl
deeepeshthakur: I did that and the order was still approx 0.5.

deeepeshthakur: [Weaved test results 1](https://htmlpreview.github.io/?https://github.com/deeepeshthakur/Urls/blob/master/12.html)
[Weaved test results 2](https://htmlpreview.github.io/?https://github.com/deeepeshthakur/Urls/blob/master/13.html)

deeepeshthakur:
@ChrisRackauckas It's working now. I'll just formalise it and push.But I have some doubt about how it's working.

src/caches/basic_method_caches.jl
```diff
@@ -115,3 +115,111 @@ function alg_cache(alg::RKMilCommute,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_
   Kj = zero(u); Dgj = zero(noise_rate_prototype)
   RKMilCommuteCache(u,uprev,du1,du2,K,gtmp,L,I,Dg,mil_correction,Kj,Dgj,tmp)
 end
+
+struct WangLi3SMil_AConstantCache <: StochasticDiffEqConstantCache end
```
ChrisRackauckas: these should all get their own file.

src/perform_step/low_order.jl
ChrisRackauckas: but you're not writing what they are doing in the paper since you're doing a derivative approximation. We could require that the user pass gg' (http://docs.juliadiffeq.org/latest/solvers/sde_solve.html#Derivative-Based-Methods-1) but I'm telling you what needs to be done for the 1st order approximation of that

src/perform_step/low_order.jl
ChrisRackauckas: Can you push the code for that?

src/caches/basic_method_caches.jl
deeepeshthakur: Will do. And If you could check whether I covered all the convergence tests?

ChrisRackauckas: what you have should be good.

src/perform_step/low_order.jl
deeepeshthakur: Oh this issue was resolved. Sorry I should have mentioned that.

ChrisRackauckas: I still see it in the code?

src/caches/basic_method_caches.jl
deeepeshthakur: I've moved these methods to a new file and I've added a few print statements because travis for rerporting timeouts.

ChrisRackauckas:
Split the test groupings so they are shorter.

PR #152, opened 2019-05-31
Fix DifferentialEquations warning

kanav99:


kanav99:
AppVeyor tests will fail most likely due to time limit, So I will merge this.

Issue #73, opened 2018-04-13
SROCK Methods


ChrisRackauckas:
- [x] SROCK2 https://hal.archives-ouvertes.fr/hal-00739754v1/document
- [x] SROCK1 https://epubs.siam.org/doi/abs/10.1137/070679375
- [x] SROCK1 Ito https://projecteuclid.org/download/pdf_1/euclid.cms/1229619673
- [x] SKSROCK https://arxiv.org/abs/1708.08145
- [x] KomBurSROCK2 https://www.sciencedirect.com/science/article/pii/S0377042712000441
- [x] TangXiaoSROCK2 https://link.springer.com/article/10.1007/s11075-018-0615-y
- [ ] SROCKC2 https://ijnao.um.ac.ir/index.php/math/article/view/69454

ChrisRackauckas:
https://arxiv.org/abs/1708.08145

ChrisRackauckas:
https://ijnao.um.ac.ir/index.php/math/article/view/69454

ChrisRackauckas:
https://www.sciencedirect.com/science/article/pii/S0377042712000441

ChrisRackauckas:
TangXiaoSROCK2 is converging with weak order 1. Could be an issue with the paper because the coefficients don't match Theorem 2's requirement.

deeepeshthakur:
@ChrisRackauckas  I think we can close this issue now.

ChrisRackauckas:
🎉

PR #153, opened 2019-06-04
Implemented SROCK method for Stratonovich form

deeepeshthakur:
@ChrisRackauckas 
Implemented SROCK method from [S-ROCK: Chebyshev Methods for Stiff Stochastic Differential](https://doi.org/10.1137/070679375)
I might have not included all the tests.
[Weaved test results](https://htmlpreview.github.io/?https://github.com/deeepeshthakur/Urls/blob/master/SROCK1_convergence_test.html)

Though there are two things that need to be finalised.
1.  Spectral Radius is assumed to be twice of that calculated by `integrator.f` only.
2.  Constant factor of the stability domain is taken to be the minimum. 

#73


ChrisRackauckas:
wrong link for the Weave

deeepeshthakur:
I forgot to add the link.Adding it. Sorry

deeepeshthakur:
> wrong link for the Weave

Done!

ChrisRackauckas:
It is shown in [3] that the S-ROCK methods applied to (3.30)
have order 1/2 in the strong sense and order 1 in the weak sense. However, if the
diffusion terms satisfy a commutativity condition
then the S-ROCK methods have strong order 1

We should make sure to document the order condition exactly like that.

ChrisRackauckas:
Convergence plots look good

ChrisRackauckas:
I didn't realize SROCK was Stratanovich convergent. 

https://math.stackexchange.com/questions/2296945/conversion-between-solution-to-stratonovich-sde-and-it%C3%B4-sde

A followup PR could do that ggprime RK calculation approximation to make a Ito version of it.

ChrisRackauckas:
>Spectral Radius is assumed to be twice of that calculated by integrator.f only.

What do you mean? Is this in the paper?

deeepeshthakur:
You know. I realised it today.For the last two days I'm testing this on Ito's form problems

deeepeshthakur:
> > Spectral Radius is assumed to be twice of that calculated by integrator.f only.
> 
> What do you mean? Is this in the paper?

Actually no. In the paper the have said to take the least number of stages such that the stability domain is greater than spectral radius

ChrisRackauckas:
https://projecteuclid.org/download/pdf_1/euclid.cms/1229619673 is the Ito one. I think we should have that be a choice that defaults to Ito, but do that in a followup PR.

deeepeshthakur:
image
This is for the second point

deeepeshthakur:
Current implementation is conservative. And assumes the smallest stabilty domain.

deeepeshthakur:
> https://projecteuclid.org/download/pdf_1/euclid.cms/1229619673 is the Ito one. I think we should have that be a choice that defaults to Ito, but do that in a followup PR.

This paper is `SROCK4`. I mean fourth in order of implementation.

ChrisRackauckas:
>This paper is SROCK4. I mean fourth in order of implementation.

I think down the line we want `SROCK(interpretation=:Stratonovich)` and `SROCK(interpretation=:Ito)`, matching the other methods, defaulting to Ito, using that paper. For now let's skip it though. Open an issue on that.

deeepeshthakur:
@ChrisRackauckas 
Travis?

PR #159, opened 2019-06-06
Optimum stage number for SROCK_1

deeepeshthakur:
@ChrisRackauckas


ChrisRackauckas:
unrelated test failure.

PR #164, opened 2019-06-08
RKMilCommute bug

deeepeshthakur:
@ChrisRackauckas


ChrisRackauckas:
Scary that no test found this. We need a case with more Brownians than SDEs

deeepeshthakur:
I think this is because we only have one test and there length(dW) ==length(uprev)

ChrisRackauckas:
Let's change that test

Issue #167, opened 2019-06-13
GPU compatibility of non-diagonal approximations


ChrisRackauckas:
https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/caches/basic_method_caches.jl#L113-L114

These should be something like `false .* ΔW.W .* ΔW.W'`. Then the iterated integral approximations need to be done via a vectorized call instead of indexing (like this).

deeepeshthakur:
Will get this done RN.

PR #169, opened 2019-06-13
RKMilCommute GPU compatibility

deeepeshthakur:
@ChrisRackauckas 
This looks good and is passing the tests
#167

### DIFF
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -96,7 +96,7 @@ struct RKMilCommuteConstantCache <: StochasticDiffEqConstantCache end
   K::rateType
   gtmp::rateNoiseType
   L::rateNoiseType
-  I::WikType
+  J::WikType
   Dg::WikType
   mil_correction::rateType
   Kj::uType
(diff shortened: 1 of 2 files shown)


src/perform_step/low_order.jl
```diff
@@ -188,18 +188,17 @@ end
...
   ggprime_norm = 0.0
 
   @.. mil_correction = zero(u)
-  for i=1:length(dW),j=1:length(dW)
-      I[j,i] = 0.5*dW[i]*dW[j]
-      if alg_interpretation(integrator.alg) == :Ito
-        j == i && (I[i,i] -= 0.5*dt) # Ito correction
-      end
+  if alg_interpretation(integrator.alg) == :Ito
+    J = 0.5 .* (vec(dW) .* vec(dW)' - dt .* I)
```
ChrisRackauckas: This is allocating, and doesn't properly fuse because of the missing `.` on the `-`.

ChrisRackauckas:
```suggestion
    J .= 0.5 .* (vec(dW) .* vec(dW)' .- dt .* I)
```

src/perform_step/low_order.jl
```diff
@@ -188,18 +188,17 @@ end
...
+  else
+    J = 0.5 .* vec(dW) .* vec(dW)'
```
ChrisRackauckas:
```suggestion
    J .= 0.5 .* vec(dW) .* vec(dW)'
```

src/caches/basic_method_caches.jl
```diff
@@ -110,8 +110,9 @@ function alg_cache(alg::RKMilCommute,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_
   du1 = zero(rate_prototype); du2 = zero(rate_prototype)
   K = zero(rate_prototype); gtmp = zero(noise_rate_prototype);
   L = zero(noise_rate_prototype); tmp = zero(rate_prototype)
-  I = zeros(length(ΔW),length(ΔW));
-  Dg = zeros(length(ΔW),length(ΔW)); mil_correction = zero(rate_prototype)
+  J = false .* vec(ΔW) .* vec(ΔW)'
```
ChrisRackauckas: If it's J then it will clash with the common name for the Jacobian. Make it WikJ?

deeepeshthakur: Okay!

src/perform_step/low_order.jl
```diff
@@ -196,7 +196,7 @@ end
 
   @.. mil_correction = zero(u)
   if alg_interpretation(integrator.alg) == :Ito
-    WikJ .= 0.5 .* (vec(dW) .* vec(dW)' .- dt .* I)
+    WikJ .= 0.5 .* (vec(dW) .* vec(dW)'  .- dt .* Diagonal(fill(one(eltype(W.dW)),length(W.dW))))
```
ChrisRackauckas: That's going to allocate. Use Eye from FillArrays.jl

ChrisRackauckas:
Going to merge, but the performance enhancement should come in a follow up PR

src/perform_step/low_order.jl
deeepeshthakur: Will check this out!

Issue #171, opened 2019-06-13
RKMilCommute allocates identity matrix, should use FillArrays


ChrisRackauckas:

deeepeshthakur:
This can be closed now, I think.

PR #173, opened 2019-06-13
RKMilCommute_Allocation

deeepeshthakur:
@ChrisRackauckas 
#169 #171

### DIFF
--- a/src/StochasticDiffEq.jl
+++ b/src/StochasticDiffEq.jl
@@ -20,7 +20,7 @@ module StochasticDiffEq
 
   using Logging
 
-  using LinearAlgebra, Random
+  using LinearAlgebra, Random, FillArrays
 
   import ForwardDiff.Dual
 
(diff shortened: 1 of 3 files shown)


src/StochasticDiffEq.jl
```diff
@@ -20,7 +20,7 @@ module StochasticDiffEq
 
   using Logging
 
-  using LinearAlgebra, Random
+  using LinearAlgebra, Random, FillArrays
```
ChrisRackauckas: Need to add FillArrays.jl to the Project.toml file.

deeepeshthakur: But Eye calls Diagonal. I not sure this will fix it.

ChrisRackauckas: `Diagonal(n)` is not a base dispatch. It's a special FillArray thing that builds the representation without building the arrays of the matrix.

deeepeshthakur: Oh. Nice!

PR #175, opened 2019-06-14
Commutative Tests

deeepeshthakur:
@ChrisRackauckas

### DIFF
--- a/test/commutative_tests.jl
+++ b/test/commutative_tests.jl
@@ -15,26 +15,30 @@ function f_commute(du,u,p,t)
 end
 
 function f_commute_analytic(u0,p,t,W)
- tmp = (A+1.01I-(B^2))*t + B*sum(W)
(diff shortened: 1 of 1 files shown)


test/commutative_tests.jl
```diff
@@ -15,26 +15,30 @@ function f_commute(du,u,p,t)
 end
 
 function f_commute_analytic(u0,p,t,W)
- tmp = (A+1.01I-(B^2))*t + B*sum(W)
+ tmp = (A+1.01I-2*(B^2))*t + B*sum(W)
```
ChrisRackauckas: what's the 2 for? I forget the analytical solution derivation here. It's from one of Platen's papers.

deeepeshthakur: that is because we have 4 noise processes.
image
We have a equation of this form and length of dW = 4

deeepeshthakur:
Travis error ???

PR #192, opened 2019-07-03
SROCK Vector Format

deeepeshthakur:
@ChrisRackauckas  
`SROCK1` using vector format only. If this passes we can use this for other SROCK methods.


src/perform_step/SROCK_perform_step.jl
```diff
@@ -146,18 +136,18 @@ end
         if typeof(W.dW) <: Number || is_diagonal_noise(integrator.sol.prob)
           @.. u += α*gₘ₋₂*W.dW
         else
-          for j in 1:length(W.dW)
-            @.. u += (α*W.dW[j])*@view(gₘ₋₂[:,j])
-          end
+          matmul!(k,gₘ₋₂,W.dW)
```
ChrisRackauckas: `mul!`?

ChrisRackauckas:
So this makes them support scalar and diagonal noise? Is there supposed to be support for non-diagonal noise as well?

src/perform_step/SROCK_perform_step.jl
deeepeshthakur: Oops!

deeepeshthakur:
It already supports non-diagonal noise

ChrisRackauckas:
oh wait I see. I read it wrong. Yeah, that's how it's done.

deeepeshthakur:
So I'll get all the other SROCK methods in this format! Don't merge it just yet.

deeepeshthakur:
@ChrisRackauckas If this passes, this is done.
Also I've not done this for TangXiaoSROCK2 because they are not functional at all. I also think we should comment out their tests.

src/SROCK_utils.jl
```diff
@@ -34,7 +34,7 @@ function maxeig!(integrator, cache::StochasticDiffEqConstantCache)
     z *= quot
   else
     dz_u = pert
-    z = dz_u*ones(z)
+    z = dz_u .* (false .* z .+ one(eltype(z)))
```
ChrisRackauckas: why is it like this? `z = dz_u*one(z)`?

deeepeshthakur:
```
julia> z = rand(4,5)
4×5 Array{Float64,2}:
 0.0543206  0.0619733  0.888743  0.822685  0.417369
 0.0877441  0.67813    0.704411  0.493939  0.239282
 0.444318   0.78979    0.3604    0.586677  0.402327
 0.344372   0.710735   0.400021  0.646333  0.178213

julia> ones(z)
ERROR: MethodError: no method matching ones(::Array{Float64,2})
Closest candidates are:
  ones(::Union{Integer, AbstractUnitRange}...) at array.jl:448
  ones(::Type{T}, ::Union{Integer, AbstractUnitRange}...) where T at array.jl:449
  ones(::Tuple{Vararg{Union{Integer, AbstractUnitRange},N} where N}) at array.jl:450
  ...
Stacktrace:
 [1] top-level scope at none:0

```

deeepeshthakur:
This time it should pass . 'TangXiaoSROCK2' was giving error. I've commented out it's tests because it's incomplete.

deeepeshthakur:
@ChrisRackauckas I think this is good to go. Passing travis.

src/SROCK_utils.jl
ChrisRackauckas: Yes, using `one(z)` would work though.

deeepeshthakur: `one(z)` gives the identity matrix. We want one at all places.

deeepeshthakur: And also the matrix `z` might not be square.

ChrisRackauckas: oh oops, yeah that's a v1.0 thing.

deeepeshthakur:
Travis is passing it's not showing here. No idea why.
image

PR #194, opened 2019-07-03
Changes for NLSolver change

kanav99:
Ref: JuliaDiffEq/DiffEqBase.jl#250

Let the tests pass first. I have some doubt too - Do SKenCarp need two copies of `linsolve` or `dz` or `jac_config`? Currently we are making two copies of these variables.

### DIFF
--- a/src/StochasticDiffEq.jl
+++ b/src/StochasticDiffEq.jl
@@ -39,13 +39,15 @@ module StochasticDiffEq
 
   using DiffEqBase: check_error!, is_diagonal_noise, @..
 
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
It shouldn't need two copies.

kanav99:
The `pr` test timed out but `push` passed. We need to split the tests even more. Should we merge this?

devmotion:
Maybe getting rid of the deprecations in DiffEqDevTools could speed up things a bit. But probably it should be adjusted at least for Appveyor.

ChrisRackauckas:
yeah devtools needs a tag I think.

PR #201, opened 2019-07-19
Fix gamma

ChrisRackauckas:


ChrisRackauckas:
@kanav99 is this correct? I forget whether you checked it.

PR #209, opened 2019-07-28
WIP Iterated Integrals

deeepeshthakur:
Implementation of Stochastic Iterated Integrals approximations from [SDELab](https://doi.org/10.1016/j.cam.2006.05.037)

### DIFF
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -117,7 +117,7 @@ function alg_cache(alg::RKMilCommute,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_
   RKMilCommuteCache(u,uprev,du1,du2,K,gtmp,L,WikJ,Dg,mil_correction,Kj,Dgj,tmp)
 end
 
-struct RKMil_GeneralConstantCache{Wiktype} <: StochasticDiffEqConstantCache
+struct RKMil_GeneralConstantCache{WikType} <: StochasticDiffEqConstantCache
   WikJ::WikType
 end
 
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -135,15 +135,19 @@ end
 end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
-  m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
-  WikJ = fill_WikJ(ΔW,Val(m),Val(false))
-  # if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-  #   WikJ = WikJDiagonal_oop(ΔW,Val(1))
-  # elseif alg.is_commutative
-  #   WikJ = WikJCommute_oop(ΔW,Val(2))
-  # else
-  #   WikJ = WikJGeneral_oop(ΔW,Val(3))
-  # end
+  # m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
+  # WikJ = fill_WikJ(ΔW,Val(m),Val(false))
+
+  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
+    # WikJ = WikJDiagonal_oop(ΔW)
+    WikJ = fill_WikJ(ΔW,Val(1),Val(false))
+  elseif alg.is_commutative
+    # WikJ = WikJCommute_oop(ΔW)
+    WikJ = fill_WikJ(ΔW,Val(2),Val(false))
+  else
+    # WikJ = WikJGeneral_oop(ΔW)
+    WikJ = fill_WikJ(ΔW,Val(3),Val(false))
+  end
   RKMil_GeneralConstantCache{typeof(WikJ)}(WikJ)
 end
 
@@ -155,15 +159,18 @@ function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate
   L = zero(noise_rate_prototype)
   mil_correction = zero(u)
   ggprime = zero(noise_rate_prototype)
-  m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
-  WikJ = fill_WikJ(ΔW,Val(m),Val(true))
-  #
-  # if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-  #   WikJ = WikJDiagonal_iip(ΔW,Val(1))
-  # elseif alg.is_commutative
-  #   WikJ = WikJCommute_iip(ΔW,Val(2))
-  # else
-  #   WikJ = WikJGeneral_iip(ΔW,Val(3))
-  # end
+  # m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
+  # WikJ = fill_WikJ(ΔW,Val(m),Val(true))
+
+  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
+    # WikJ = WikJDiagonal_iip(ΔW)
+    WikJ = fill_WikJ(ΔW,Val(1),Val(true))
+  elseif alg.is_commutative
+    # WikJ = WikJCommute_iip(ΔW)
+    WikJ = fill_WikJ(ΔW,Val(2),Val(true))
+  else
+    # WikJ = WikJGeneral_iip(ΔW)
+    WikJ = fill_WikJ(ΔW,Val(3),Val(true))
+  end
   RKMil_GeneralCache{typeof(u), typeof(rate_prototype), typeof(noise_rate_prototype), typeof(WikJ)}(u, uprev, tmp, du₁, du₂, K, L, mil_correction, ggprime, WikJ)
 end
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -117,8 +117,8 @@ function alg_cache(alg::RKMilCommute,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_
   RKMilCommuteCache(u,uprev,du1,du2,K,gtmp,L,WikJ,Dg,mil_correction,Kj,Dgj,tmp)
 end
 
-struct RKMil_GeneralConstantCache <: StochasticDiffEqConstantCache
-  m_seq::Array{Int}
+struct RKMil_GeneralConstantCache{Wiktype} <: StochasticDiffEqConstantCache
+  WikJ::WikType
 end
 
 @cache struct RKMil_GeneralCache{uType, rateType, rateNoiseType, WikType} <: StochasticDiffEqMutableCache
@@ -132,31 +132,17 @@ end
   mil_correction::uType
   ggprime::rateNoiseType
   WikJ::WikType
-  WikJ2::WikType
-  WikJ3::WikType
-  m_seq::Array{Int}
-  vec_ζ::Vector{eltype(rateNoiseType)}
-  vec_η::Vector{eltype(rateNoiseType)}
-  Gp1::Vector{eltype(rateNoiseType)}
-  Gp2::Vector{eltype(rateNoiseType)}
 end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
-  if typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative
-    m_seq = zeros(Int,length(ΔW))
+  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
+    WikJ = fill_WikJDiagonal_oop(ΔW)
+  elseif alg.is_commutative
+    WikJ = fill_WikJCommute_oop(ΔW)
   else
-    M = length(ΔW) * (length(ΔW) - 1)/2
-    m_seq = Array{Int}(undef, M, 2)
-    k = 1
-    for i in 1:length(ΔW)
-      for j in i+1:length(ΔW)
-        m_seq[k,1] = i
-        m_seq[k,2] = j
-        k += 1
-      end
-    end
+    WikJ = fill_WikJGeneral_oop(ΔW)
   end
-  RKMil_GeneralConstantCache(m_seq)
+  RKMil_GeneralConstantCache{typeof(WikJ)}(WikJ)
 end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{true}})
@@ -168,40 +154,11 @@ function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate
   mil_correction = zero(u)
   ggprime = zero(noise_rate_prototype)
   if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-    WikJ = false .* ΔW .* ΔW
+    WikJ = fill_WikJDiagonal_iip(ΔW)
+  elseif alg.is_commutative
+    WikJ = fill_WikJCommute_iip(ΔW)
   else
-    WikJ = false .* vec(ΔW) .* vec(ΔW)'
+    WikJ = fill_WikJGeneral_iip(ΔW)
   end
-
-  #in case of diagonal noise and commutative noise
-  # WikJ2, WikJ3, vec_η, Gp1, Gp2 are dummied variables
-  if typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative
-    m_seq = zeros(Int,length(ΔW))
-    WikJ2 = WikJ
-    WikJ3 = WikJ
-    vec_ζ = false .* vec(ΔW)
-    vec_η = vec_ζ
-    Gp1 = vec_ζ
-    Gp2 = vec_ζ
-  else
-    M = length(ΔW) * (length(ΔW) - 1)/2
-    m_seq = Array{Int}(undef, M, 2)
-    k = 1
-    for i in 1:length(ΔW)
-      for j in i+1:length(ΔW)
-        m_seq[k,1] = i
-        m_seq[k,2] = j
-        k += 1
-      end
-    end
-    WikJ2 = similar(WikJ)
-    WikJ3 = similar(WikJ)
-    vec_ζ = false .* vec(ΔW)
-    vec_η = similar(vec_ζ)
-    Gp1 = zeros(M)
-    Gp2 = similar(Gp1)
-  end
-
-  RKMil_GeneralCache{typeof(u), typeof(rate_prototype), typeof(noise_rate_prototype), typeof(WikJ)}(u, uprev, tmp, du₁,
-                                du₂, K, L, mil_correction, ggprime, WikJ, WikJ2, WikJ3, m_seq, vec_ζ, vec_η, Gp1, Gp2)
+  RKMil_GeneralCache{typeof(u), typeof(rate_prototype), typeof(noise_rate_prototype), typeof(WikJ)}(u, uprev, tmp, du₁, du₂, K, L, mil_correction, ggprime, WikJ)
 end
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -135,13 +135,15 @@ end
 end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
-  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-    WikJ = WikJDiagonal_oop(ΔW)
-  elseif alg.is_commutative
-    WikJ = WikJCommute_oop(ΔW)
-  else
-    WikJ = WikJGeneral_oop(ΔW)
-  end
+  m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
+  WikJ = fill_WikJ(ΔW,Val(m),Val(false))
+  # if typeof(ΔW) <: Number || is_diagonal_noise(prob)
+  #   WikJ = WikJDiagonal_oop(ΔW,Val(1))
+  # elseif alg.is_commutative
+  #   WikJ = WikJCommute_oop(ΔW,Val(2))
+  # else
+  #   WikJ = WikJGeneral_oop(ΔW,Val(3))
+  # end
   RKMil_GeneralConstantCache{typeof(WikJ)}(WikJ)
 end
 
@@ -153,12 +155,15 @@ function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate
   L = zero(noise_rate_prototype)
   mil_correction = zero(u)
   ggprime = zero(noise_rate_prototype)
-  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-    WikJ = WikJDiagonal_iip(ΔW)
-  elseif alg.is_commutative
-    WikJ = WikJCommute_iip(ΔW)
-  else
-    WikJ = WikJGeneral_iip(ΔW)
-  end
+  m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
+  WikJ = fill_WikJ(ΔW,Val(m),Val(true))
+  #
+  # if typeof(ΔW) <: Number || is_diagonal_noise(prob)
+  #   WikJ = WikJDiagonal_iip(ΔW,Val(1))
+  # elseif alg.is_commutative
+  #   WikJ = WikJCommute_iip(ΔW,Val(2))
+  # else
+  #   WikJ = WikJGeneral_iip(ΔW,Val(3))
+  # end
   RKMil_GeneralCache{typeof(u), typeof(rate_prototype), typeof(noise_rate_prototype), typeof(WikJ)}(u, uprev, tmp, du₁, du₂, K, L, mil_correction, ggprime, WikJ)
 end
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -117,8 +117,7 @@ function alg_cache(alg::RKMilCommute,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_
   RKMilCommuteCache(u,uprev,du1,du2,K,gtmp,L,WikJ,Dg,mil_correction,Kj,Dgj,tmp)
 end
 
-struct RKMil_GeneralConstantCache{WikType} <: StochasticDiffEqConstantCache
-  WikJ::WikType
+struct RKMil_GeneralConstantCache <: StochasticDiffEqConstantCache
   m_seq::Array{Int}
 end
 
@@ -143,12 +142,6 @@ end
 end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
-  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-    WikJ = false .* ΔW .* ΔW
-  else
-    WikJ = false .* vec(ΔW) .* vec(ΔW)'
-  end
-
   if typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative
     m_seq = zeros(Int,length(ΔW))
   else
@@ -163,7 +156,7 @@ function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate
       end
     end
   end
-  RKMil_GeneralConstantCache{typeof(WikJ)}(WikJ, m_seq)
+  RKMil_GeneralConstantCache(m_seq)
 end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{true}})
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -135,19 +135,7 @@ end
 end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
-  # m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
-  # WikJ = fill_WikJ(ΔW,Val(m),Val(false))
-
-  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-    # WikJ = WikJDiagonal_oop(ΔW)
-    WikJ = fill_WikJ(ΔW,Val(1),Val(false))
-  elseif alg.is_commutative
-    # WikJ = WikJCommute_oop(ΔW)
-    WikJ = fill_WikJ(ΔW,Val(2),Val(false))
-  else
-    # WikJ = WikJGeneral_oop(ΔW)
-    WikJ = fill_WikJ(ΔW,Val(3),Val(false))
-  end
+  WikJ = get_WikJ(ΔW,prob,alg)
   RKMil_GeneralConstantCache{typeof(WikJ)}(WikJ)
 end
 
@@ -159,18 +147,6 @@ function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate
   L = zero(noise_rate_prototype)
   mil_correction = zero(u)
   ggprime = zero(noise_rate_prototype)
-  # m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
-  # WikJ = fill_WikJ(ΔW,Val(m),Val(true))
-
-  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-    # WikJ = WikJDiagonal_iip(ΔW)
-    WikJ = fill_WikJ(ΔW,Val(1),Val(true))
-  elseif alg.is_commutative
-    # WikJ = WikJCommute_iip(ΔW)
-    WikJ = fill_WikJ(ΔW,Val(2),Val(true))
-  else
-    # WikJ = WikJGeneral_iip(ΔW)
-    WikJ = fill_WikJ(ΔW,Val(3),Val(true))
-  end
+  WikJ = get_WikJ(ΔW,prob,alg)
   RKMil_GeneralCache{typeof(u), typeof(rate_prototype), typeof(noise_rate_prototype), typeof(WikJ)}(u, uprev, tmp, du₁, du₂, K, L, mil_correction, ggprime, WikJ)
 end
--- a/src/caches/basic_method_caches.jl
+++ b/src/caches/basic_method_caches.jl
@@ -136,11 +136,11 @@ end
 
 function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
   if typeof(ΔW) <: Number || is_diagonal_noise(prob)
-    WikJ = fill_WikJDiagonal_oop(ΔW)
+    WikJ = WikJDiagonal_oop(ΔW)
   elseif alg.is_commutative
-    WikJ = fill_WikJCommute_oop(ΔW)
+    WikJ = WikJCommute_oop(ΔW)
   else
-    WikJ = fill_WikJGeneral_oop(ΔW)
+    WikJ = WikJGeneral_oop(ΔW)
   end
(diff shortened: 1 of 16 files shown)


ChrisRackauckas:
You mean SDELab2?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,130 @@
...
+    @unpack dt, u, uprev, t, p, W = integrator
+    @unpack m_seq, WikJ = cache
+
+    dW     = W.dW
+    m      = length(dW)
+    M      = m*(m-1)/2
+
+    if typef(dW) <: Number || is_diagonal_noise(integrator.sol.prob)
+        cache.WikJ = 1//2 .* J .^ 2
+    else
```
ChrisRackauckas: what about the commutative noise case? That should be in the RKMilCommute cache constructor already.

deeepeshthakur:
> You mean SDELab2?

SDELab2 had a few mistakes, so I tried sticking to the paper.

src/iterated_integrals.jl
deeepeshthakur: We don't have a function for commutative noise like we have for diagonal so I thought let's consider it in general case. Otherwise, It would be difficult maintaining a commutative noise version of every solver.

ChrisRackauckas: That's what this is: https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/perform_step/low_order.jl#L198-L202 . It should just call to a common routine in here to fill `WikJ` instead, and if it's commutative that's the calculation.

ChrisRackauckas:
Oh okay. Was just checking for licensing reasons. If it's from the paper description then you're good.

deeepeshthakur:
IIRC SDELAB2 is also MIT license😋

deeepeshthakur:
@ChrisRackauckas 
Right now what I've done is that I've added an additional method `RKMil_General` that can handle 1-D, Diagonal, Commutative, and General M-Dimensional noise. I'm testing it. If it passes the tests then we can replace the other `RKMil` methods with this in a different PR.
Also for commutative noise, I've added an option in alg part.

src/iterated_integrals.jl
```diff
@@ -15,9 +15,6 @@ function get_iterated_I!(integrator, cache::StochasticDiffEqConstantCache)
         m_seq  = cache.m_seq
         sum_dW² = zero(eltype(dW))
```
devmotion: That line is not needed it seems?

deeepeshthakur: Yes. Missed that. Thank You

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,124 @@ end
 
   end
 end
+
+@muladd function perform_step!(integrator,cache::RKMil_GeneralConstantCache,f=integrator.f)
+  @unpack t,dt,uprev,u,W,p = integrator
+  dW = W.dW
+
+  if integrator.alg.is_commutative
+    WikJ = 1//2 .* vec(dW) .* vec(dW)'
```
ChrisRackauckas: I think this could be handled as a branch in the iterated I just like diagonal?

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,124 @@ end
...
+  else
+    WikJ = get_iterated_I!(integrator, cache)
```
ChrisRackauckas: We should probably have a struct for WikJDiagonal, WikJCommute, WikJGeneral that dispatches to different ways of filling it? Then it could have the appropriate cache size for each case, and the cache variables won't necessarily need to be in the method cache (allowing for simultaneous different WikJ's in future methods)

src/perform_step/low_order.jl
deeepeshthakur: But there we have a function for diagonal noise and the commutative noise it's alg specific

ChrisRackauckas: The way to generate WikJ for commutative noise isn't alg-specific.

deeepeshthakur: No not that. I mean we don't have an efficient way to check for commutative noise like we have for diagonal noise so rn commutative noise is just an option in the alg struct

ChrisRackauckas: I see. Make it a value type so that we can dispatch on it and do a cleaner WikJ construction, and we should find a better system for noise classifications later.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    WikJ = false .* ΔW .* ΔW'
+    WikJCommute_oop{typeof(WikJ)}(WikJ)
+end
+
+function fill_WikJCommute_iip(ΔW)
+    WikJ = false .* ΔW .* ΔW'
+    WikJCommute_iip{typeof(WikJ)}(WikJ)
+end
+
+function fill_WikJGeneral_oop(ΔW)
```
ChrisRackauckas: fill_WikJCommute_oop?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    for i in 1:length(ΔW)
+      for j in i+1:length(ΔW)
+        m_seq[k,1] = i
+        m_seq[k,2] = j
+        k += 1
+      end
+    end
+    vec_ζ = false .* vec(ΔW)
+    vec_η = similar(vec_ζ)
+    Gp1 = zeros(M)
```
ChrisRackauckas: Doing this because you want the lower triangle only? It's probably better to just make the matrix so that this can be GPU-friendly.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    WikJ
+end
+
+function get_iterated_I!(dW, Wik::WikJDiagonal_iip)
+    @unpack WikJ = Wik
+    if typeof(dW) <: Number
+        Wik.WikJ = 1//2 .* dW .^ 2
+    else
+        @.. WikJ = 1//2*dW^2
+        @.. Wik.WikJ = WikJ
```
ChrisRackauckas: this line isn't needed?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+function get_iterated_I!(dW, Wik::WikJDiagonal_oop)
+    @unpack WikJ = Wik
+    WikJ = 1//2 .* dW .* dW
+    WikJ
+end
+
+function get_iterated_I!(dW, Wik::WikJDiagonal_iip)
+    @unpack WikJ = Wik
+    if typeof(dW) <: Number
+        Wik.WikJ = 1//2 .* dW .^ 2
```
ChrisRackauckas: this works whether dW is a number or an array?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    WikJ = 1//2 .* vec(dW) .* vec(dW)'
+    WikJ
+end
+
+function get_iterated_I!(dW, Wik::WikJCommute_iip)
+    @unpack WikJ = Wik
+    mul!(WikJ,vec(dW),vec(dW)')
+    @.. WikJ *= 1//2
+
+    @.. Wik.WikJ = WikJ
```
ChrisRackauckas: this isn't needed? WikJ === Wik.WikJ

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,127 @@ end
...
+  Wik = cache.WikJ
+  dW = W.dW
+
+  WikJ = get_iterated_I!(dW, Wik)
+
+  if alg_interpretation(integrator.alg) == :Ito
+    if typeof(dW) <: Number || is_diagonal_noise(integrator.sol.prob)
+      WikJ = WikJ .- 1//2 .* dt
+    else
+      WikJ -= 1//2 .* dt .* Eye{eltype(dW)}(length(dW))
```
ChrisRackauckas: Are you sure about the identity matrix here?

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,127 @@ end
...
+  @unpack t,dt,uprev,u,W,p = integrator
+  dW = W.dW; sqdt = integrator.sqdt
+  Wik = cache.WikJ
+  get_iterated_I!(dW, Wik)
+  WikJ = Wik.WikJ
+
+  integrator.f(du₁,uprev,p,t)
+  integrator.g(L,uprev,p,t)
+  @.. mil_correction = zero(eltype(u))
+  ggprime_norm = zero(eltype(ggprime))
```
ChrisRackauckas: just do `local ggprime_norm`

test/sde/sde_convergence_tests.jl
```diff
@@ -22,6 +22,8 @@ sim  = test_convergence(dts,prob,LambaEM(),trajectories=Int(1e2))
 @test abs(sim.𝒪est[:l2]-.5) < 0.2
```
ChrisRackauckas: we need a non-commutative convergence test now. Does https://epubs.siam.org/doi/abs/10.1137/09076636X have one? IIRC it does

ChrisRackauckas:
If you use RKMil_General and RKMil on the same random seed do they compute exactly the same values, with the same timestepping, when it's diagonal noise? Same with RKMilCommute? I think both of those algorithms can be subsumed, or at least the former can be renamed (for testing) and this can be made `RKMil`. From a quick look, it seems that the caching is done correctly so it will be just as efficient in that case as well, and the error estimator should be exactly the same.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    return nothing
+end
+
```
ChrisRackauckas: Comment or docstring that points to where this algorithm comes from.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+        Wik.WikJ = 1//2 .* dW .^ 2
+    else
+        @.. WikJ = 1//2*dW^2
+        @.. Wik.WikJ = WikJ
+    end
+    return nothing
+end
+
+function get_iterated_I!(dW, Wik::WikJCommute_oop)
+    @unpack WikJ = Wik
```
ChrisRackauckas: why have a cached version at all then? This Wik type should probably have no fields

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    m_seq = Array{Int}(undef, M, 2)
+    k = 1
+    for i in 1:length(ΔW)
+      for j in i+1:length(ΔW)
+        m_seq[k,1] = i
+        m_seq[k,2] = j
+        k += 1
+      end
+    end
+    WikJCommute_oop{eltype(ΔW), typeof(WikJ)}(WikJ, m_seq)
```
ChrisRackauckas:
```suggestion
    WikJGeneral_oop{eltype(ΔW), typeof(WikJ)}(WikJ, m_seq)
```

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+
+    #operator (Iₘ² - Pₘ)Kₘᵀ
+    for i in 1:M
+        WikJ2[m_seq[i,1], m_seq[i,2]] = Gp[i]
+        WikJ2[m_seq[i,2], m_seq[i,1]] = -Gp[i]
+    end
+
+    WikJ *= 1//2
+    a2ₚ = (π^2)/6
+    p = Int(floor((1/π)*sqrt(M/(24*dt))*sqrt(m + 4*sum_dW²/dt) + 1))
```
ChrisRackauckas: ceil? Also, this should have a tweakable parameter `C` from the algorithm

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    for i in 1:p
+        a2ₚ -= (1/i^2)
+        var = sqrt(dt/(2*π*i))
```
ChrisRackauckas: zeta has a sqrt(dt), eta has a sqrt(dt) but then in Ap you divide by dt, so this should be 1/sqrt(dt)?

ChrisRackauckas: oh nevermind, then it's multiplied by dt... that is written horribly in the paper lol

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+        vec_ζ = randn(m)*var
+        vec_η = randn(m)*var
+        WikJ += (vec_ζ*vec_η' - vec_η*vec_ζ')
+        Aₚ -= (2/sqrt(π*i))*vec_ζ
```
ChrisRackauckas: little a?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    WikJ -= 1//2*(dW*Aₚ' - Aₚ*dW')
+    WikJ += (sqrt(a2ₚ)*dt/π)*WikJ2
```
ChrisRackauckas: why is there a dt here?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    end
+    Gp = Gp/sqrt(2) + Gp2
+
+    #operator (Iₘ² - Pₘ)Kₘᵀ
+    for i in 1:M
+        WikJ2[m_seq[i,1], m_seq[i,2]] = Gp[i]
+        WikJ2[m_seq[i,2], m_seq[i,1]] = -Gp[i]
+    end
+
+    WikJ *= 1//2
```
ChrisRackauckas: for legibility, this stuff should come before the tail correction

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    a2ₚ = (π^2)/6
```
ChrisRackauckas: I don't see this in the paper at all.

src/iterated_integrals.jl
ChrisRackauckas: In fact, describe the computation that you did, like https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/nlsolve/newton.jl#L1-L43

src/iterated_integrals.jl
```diff
@@ -0,0 +1,237 @@
...
+    @unpack WikJ, m_seq = Wik
+
+    m      = length(dW)
+    M      = m*(m-1)/2
+    sum_dW² = dW'*dW
+
+    WikJ = dW*dW'
+    Gp1 = randn(M)
+    α = sqrt(1 + sum_dW²/dt)
+    Gp2 = Gp1/(sqrt(2)*(1+α)*dt)
```
ChrisRackauckas: I fail to see how this is Sigma infty?

ChrisRackauckas:
Pulling in @YingboMa as well since this is a quite important one.

src/iterated_integrals.jl
deeepeshthakur: I forgot that. Sorry.

src/iterated_integrals.jl
deeepeshthakur: Yes. So You are saying that I should just make this a normal matrix instead?

src/iterated_integrals.jl
deeepeshthakur: It was not working for a number so I made the struct mutable and added this branching.

src/perform_step/low_order.jl
deeepeshthakur: It was working for the normal RKMil_Commute, so I don't see why it should be a problem here.

src/perform_step/low_order.jl
deeepeshthakur: I'm not sure if I'm getting what you are saying here.

test/sde/sde_convergence_tests.jl
deeepeshthakur: Yes, It does. with m = 4. We should also find something similar for Stratonovich problems

deeepeshthakur:
> If you use RKMil_General and RKMil on the same random seed do they compute exactly the same values, with the same timestepping, when it's diagonal noise? Same with RKMilCommute? I think both of those algorithms can be subsumed, or at least the former can be renamed (for testing) and this can be made `RKMil`. From a quick look, it seems that the caching is done correctly so it will be just as efficient in that case as well, and the error estimator should be exactly the same.

I was thinking that this was the whole purpose of this. I mean I thought we would completely remove the other 3 methods when this gets going and rename `RKMil_General` to `RKMil`

src/iterated_integrals.jl
deeepeshthakur: This sounds nice. Because it's quite complicated even though I tried to be as close to the paper as I could.

src/iterated_integrals.jl
deeepeshthakur: This I have taken from SDELAB2, There C is taken to be 1

deeepeshthakur: I did `floor(ans + 1)`. Will change it to ceil.

src/iterated_integrals.jl
deeepeshthakur: Yes. I skipped some scalar calculations because they canceled themselves.

src/iterated_integrals.jl
deeepeshthakur: Sum(1/x^2) from 1 to inf is pi^2/6. And then I subtract 1/x^2 from 1 to p.

src/iterated_integrals.jl
deeepeshthakur: No. Actually, I've changed the order in which we multiply scalars. Sigma infty is `a_p`

src/iterated_integrals.jl
deeepeshthakur: image
This is the scalar `Delta_t`/`2*pi`*`ap^0.5`. But I skipped the division by 2 due to cancellation

src/iterated_integrals.jl
deeepeshthakur: I changed this part to vector calculations image. So `A_p` is the vector of `a_i0`.

deeepeshthakur: And I think the `i` in line 174 is r in the paper.

src/iterated_integrals.jl
ChrisRackauckas: let's keep that for a future PR. This is complex already.

src/iterated_integrals.jl
ChrisRackauckas: can you rename a_p then?

src/iterated_integrals.jl
ChrisRackauckas: he was probably being lazy with C. Make C=1 the default but it should still be a parameter.

src/iterated_integrals.jl
ChrisRackauckas: can we change the names so it lines up?

src/perform_step/low_order.jl
ChrisRackauckas: oh yes, now I remember why it's there.

src/perform_step/low_order.jl
ChrisRackauckas: instead of initializing it you can just localize it

src/iterated_integrals.jl
deeepeshthakur: I did this because of a memory optimization.

ChrisRackauckas: how does it optimize the memory?

deeepeshthakur: There We'll need one extra WikJ type array if we do that.

src/iterated_integrals.jl
deeepeshthakur: Oh sorry, you are referring to the operator sigma infinity. I broke that down into some smaller operators.

src/iterated_integrals.jl
deeepeshthakur: I've added some markdown style docs. Hope this makes it easier to understand.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,286 @@
...
+    WikJ3::WikJType
+    m_seq::Array{Int}
+    vec_ζ::Vector{eltype(rateNoiseElTypeNoUnits)}
+    vec_η::Vector{eltype(rateNoiseElTypeNoUnits)}
+    Gp1::Vector{eltype(rateNoiseElTypeNoUnits)}
+    Gp2::Vector{eltype(rateNoiseElTypeNoUnits)}
+    Aᵢ::Vector{eltype(rateNoiseElTypeNoUnits)}
+end
+
+function fill_WikJDiagonal_oop(ΔW)
```
ChrisRackauckas: This should just be the constructor for the type, i.e. have the same name as the type.

src/caches/basic_method_caches.jl
```diff
@@ -116,3 +116,49 @@ function alg_cache(alg::RKMilCommute,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_
...
+  L::rateNoiseType
+  mil_correction::uType
+  ggprime::rateNoiseType
+  WikJ::WikType
+end
+
+function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
+  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
+    WikJ = fill_WikJDiagonal_oop(ΔW)
+  elseif alg.is_commutative
```
ChrisRackauckas: since there is a type change here, this is something we want to resolve at compile time.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,286 @@
+"""
+
+    get_iterated_I!(dW, Wik::AbstractWikJ, C)
```
ChrisRackauckas: this needs to be right above `get_iterated_I!` for the docstring to work.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,286 @@
...
+Here we've only shown in which order these are implemented in this code.
+
+From above we can see:
+
+```math
+    Δt/(2π) √(𝑎ₚ) √(Σ∞) Gₚ = Δt/π √(𝑎ₚ) (√(Σ∞)/2 Gₚ)
+```
+
+```math
+    Oper1(Gₚ) = (√(Σ∞)/2 Gₚ) = (Iₘ/√2 + 1/(√2 * (1 + α) * Δt) Kₘ(Iₘ² - Pₘ)(Iₘ ⨂ W(Δt)W(Δt)ᵀ)(Iₘ² - Pₘ)Kₘᵀ)Gₚ
```
ChrisRackauckas: If this is Gp1, then call it Gp1. If it's not Gp1, what is it?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,286 @@
...
+
+    #operator (Iₘ X W*Wᵀ)
+    WikJ2 = WikJ*WikJ2
+
+    #operator Kₘ(Iₘ² - Pₘ)
+    WikJ2 = WikJ2 - WikJ2'
+    for i in 1:M
+        Gp2[i] = WikJ2[m_seq[i,1], m_seq[i,2]]
+    end
+    Gp = Gp/sqrt(2) + Gp2
```
ChrisRackauckas: Gp on the right hand side here is undefined? Also, where does this relation come from?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,286 @@
...
+    #operator (Iₘ X W*Wᵀ)
+    mul!(WikJ,dW,dW')
+    mul!(WikJ3,WikJ,WikJ2)
+
+    #operator Kₘ(Iₘ² - Pₘ)
+    @.. WikJ2 = WikJ3 - WikJ3'
+    for i in 1:M
+        Gp2[i] = WikJ2[m_seq[i,1], m_seq[i,2]]
+    end
+    Gp1 = Gp1/sqrt(2) + Gp2
```
ChrisRackauckas: @..

src/iterated_integrals.jl
deeepeshthakur: image

So what I'm actually doing is applying the Square root sigma infinity operator on `Gp`. So this can be broken down into `I_M` part and the remaining part.
So when I'm saying ` Gp1 = Gp1/sqrt(2) + Gp2` here `Gp1` is the initial `Gp1` as `I_M * Gp1` will still be `Gp` and `Gp2` is the rest of the operator applied to `Gp1`. Thus this expression will give the Sigma_Infinity Operator of `Gp1`.

src/caches/basic_method_caches.jl
deeepeshthakur: Then we'll have to write different functions for cache initializations.

ChrisRackauckas: no you won't

deeepeshthakur: How can we resolve this in compile-time without making separate functions?

ChrisRackauckas: make it type-based, like `Val(true)` or something.

src/caches/basic_method_caches.jl
```diff
@@ -116,3 +116,54 @@ function alg_cache(alg::RKMilCommute,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_
...
+  K::uType
+  L::rateNoiseType
+  mil_correction::uType
+  ggprime::rateNoiseType
+  WikJ::WikType
+end
+
+function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,f,t,dt,::Type{Val{false}})
+  m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
+  WikJ = fill_WikJ(ΔW,Val(m),Val(false))
```
deeepeshthakur: @ChrisRackauckas I tried this. But I don't see how this might have any compile-time benefits?

src/caches/basic_method_caches.jl
```diff
@@ -155,15 +159,18 @@ function alg_cache(alg::RKMil_General,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate
...
-  #   WikJ = WikJCommute_iip(ΔW,Val(2))
-  # else
-  #   WikJ = WikJGeneral_iip(ΔW,Val(3))
-  # end
+  # m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
+  # WikJ = fill_WikJ(ΔW,Val(m),Val(true))
+
+  if typeof(ΔW) <: Number || is_diagonal_noise(prob)
+    # WikJ = WikJDiagonal_iip(ΔW)
+    WikJ = fill_WikJ(ΔW,Val(1),Val(true))
```
ChrisRackauckas: ?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,302 @@
...
+Here we've only shown in which order these are implemented in this code.
+
+From above we can see:
+
+```math
+    Δt/(2π) √(𝑎ₚ) √(Σ∞) Gp₁ = Δt/π √(𝑎ₚ) (√(Σ∞)/2 Gp₁)
+```
+
+```math
+    Oper2(Gp₁) = (√(Σ∞)/2 Gp₁) = (Iₘ/√2 + F/(√2 * (1 + α) * Δt))(Gp₁)
```
ChrisRackauckas: what's Oper2?

deeepeshthakur: I just named it so that I won't have to write the whole left part again.

ChrisRackauckas: But what is it? What operation is it? Is it in the code?

deeepeshthakur: No. Just explanation purposes.

src/caches/basic_method_caches.jl
deeepeshthakur: I still don't understand the purpose of Value types here. 
Earlier what I did was
```
  m = (typeof(ΔW) <: Number || is_diagonal_noise(prob))*1 + (alg.is_commutative)*2 + (!(typeof(ΔW) <: Number || is_diagonal_noise(prob) || alg.is_commutative))*3
  WikJ = fill_WikJ(ΔW,Val(m),Val(false))
```
Got compilation error.
Then I changed it to:
```
    WikJ = WikJDiagonal_oop(ΔW,Val(true))
```
But I guess this won't make any difference because there is only 1 definition of `WikJDiagonal_oop`

So I did
```
    WikJ = fill_WikJ(ΔW,Val(1),Val(false))
```
1, 2, or 3 for Diagonal, commutative and general noise and true or false for in place and out of place.

ChrisRackauckas: no, the purpose was for `is_commutative` to be something that is a value type so it can be a compile-time check. Examples: http://www.stochasticlifestyle.com/type-dispatch-design-post-object-oriented-programming-julia/

src/iterated_integrals.jl
```diff
@@ -0,0 +1,302 @@
...
+    𝒜ᵖ = √(𝑎ₚ)*Δt/π * (Iₘ² - Pₘ)Kₘᵀ(Oper2(Gp₁))
+```
+In the code we have
+
+```math
+    WikJ2 = (Iₘ² - Pₘ)Kₘᵀ(Oper2(Gp₁))
+```
+
+"""
+
```
ChrisRackauckas: need to put this over the correct dispatch

src/iterated_integrals.jl
```diff
@@ -0,0 +1,302 @@
...
+        WikJ2[m_seq[i,1], m_seq[i,2]] = Gp₂[i]
+        WikJ2[m_seq[i,2], m_seq[i,1]] = -Gp₂[i]
+    end
+
+    #operator (Iₘ X W*Wᵀ)
+    WikJ2 = WikJ*WikJ2
+
+    #operator Kₘ(Iₘ² - Pₘ)
+    WikJ2 = WikJ2 - WikJ2'
+    for i in 1:M
```
ChrisRackauckas: this loop isn't necessary?

src/iterated_integrals.jl
```diff
@@ -0,0 +1,302 @@
...
+```
+
+initially we have
+
+    Gp₂ = Gp₁/(sqrt(2)*(1+α)*dt)
+
+thus applying operator `F` on `Gp₂`. And hence we have
+
+    Gp₂ = F(Gp₂)
+    Gp₁ = (Gp₁/√2) + Gp₂
```
ChrisRackauckas: `(1-1/sqrt(2))Gp₁ = Gp₂`? That doesn't seem right.

deeepeshthakur: No no. This is an assignment. This is `=` not `==`.

src/iterated_integrals.jl
ChrisRackauckas: nevermind

src/iterated_integrals.jl
```diff
@@ -0,0 +1,302 @@
...
+    (√(Σ∞)/2 Gp₁) = Gp₁/√2 + F(Gp₁/(√2*(1+α)*Δt))
+```
+
+initially we have
+
+    Gp₂ = Gp₁/(sqrt(2)*(1+α)*dt)
+
+thus applying operator `F` on `Gp₂`. And hence we have
+
+    Gp₂ = F(Gp₂)
```
ChrisRackauckas: F is the identity?

src/iterated_integrals.jl
deeepeshthakur: I think we need this.

src/iterated_integrals.jl
ChrisRackauckas: It shouldn't be. The whole purpose of the docstring is to be a clear unambiguous description of what is being calculated. This is precisely the issue. You shouldn't write mathematical descriptions in a way that is dependent on the order that it is read in: that's for code which has an order.

src/iterated_integrals.jl
deeepeshthakur: F is mentioned above. line 124.

ChrisRackauckas: `Gp₂ = F(Gp₂)` => `F` is the identity.

deeepeshthakur: This is the same issue as above. https://github.com/JuliaDiffEq/StochasticDiffEq.jl/pull/209#discussion_r314223449

ChrisRackauckas: Yes it is

src/caches/basic_method_caches.jl
deeepeshthakur: But then we'll have to take it as an argument. Like we can default it to diagonal noise case but then for general noise, the user will have to give another argument.

src/iterated_integrals.jl
deeepeshthakur: Fixed it. I think.

ChrisRackauckas:
I think the description is good now. I refactored the iterated integral setup code so that it ends up type stable from a val dispatch on whether to use the commutative noise approximation. 

I will say that this code is good to go, but it's still missing a stress test on the general noise case. We need a convergence test that showcases that the iterated integral approximation really is working. There is a test case in here which gives us good confidence, and thus I am willing to merge, but a convergence test would be better. Let's add that in a future PR.

deeepeshthakur:
I'll fix the undef error and typo in doc strings

src/iterated_integrals.jl
```diff
@@ -24,7 +24,7 @@ struct WikJGeneral_oop <: AbstractWikJGeneral
     m_seq::Matrix{Int}
     function WikJGeneral_oop(ΔW)
         m = length(ΔW)
-        M = m*(m-1)/2
+        M = Int(m*(m-1)/2)
```
devmotion: I'd suggest `M = div(m * (m - 1), 2)` instead. A quick check with
```julia
f(m) = div(m * (m - 1), 2); g(m) = Int(m * (m - 1) / 2)
```
shows that the output of `@code_native` and `@code_llvm` is a lot simpler for `f`.

ChrisRackauckas:
Note `RKMil_General` shouldn't be needed, since it could just be `RKMil`, which will auto-specialize to the special case of diagonal noise? We need to check that it's just as efficient there, but if it is, then this can be the one algorithm for the RKMils. RKMilCommute can stay an alias to the IICommutative version with a deprecation warning. That should be a followup PR.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,345 @@
...
+    m_seq::Matrix{Int}
+    vec_ζ::Vector{rateNoiseElTypeNoUnits}
+    vec_η::Vector{rateNoiseElTypeNoUnits}
+    Gp₁::Vector{rateNoiseElTypeNoUnits}
+    Gp₂::Vector{rateNoiseElTypeNoUnits}
+    Aᵢ::Vector{rateNoiseElTypeNoUnits}
+end
+
+function WikJGeneral_iip(ΔW)
+    WikJ = false .* ΔW .* ΔW'
```
YingboMa: `false .* vec(ΔW) .* vec(ΔW)'`

src/iterated_integrals.jl
```diff
@@ -0,0 +1,345 @@
...
+        m_seq[k,1] = i
+        m_seq[k,2] = j
+        k += 1
+      end
+    end
+    vec_ζ = false .* vec(ΔW)
+    vec_η = false .* vec(ΔW)
+    Gp₁ = false .* Array{eltype(ΔW)}(undef, M)
+    Gp₂ = false .* Array{eltype(ΔW)}(undef, M)
+    Aᵢ = false .* vec(ΔW)
```
YingboMa: Those `false .*` are unnecessary and inefficient, because the desired array has the same shape with `ΔW`, so `zero` and `fill! + similar` is sufficient.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,345 @@
...
+```math
+    WikJ2 = (Iₘ² - Pₘ)Kₘᵀ(Oper2(Gp₁))
+```
+
+"""
+function get_iterated_I!(dt, dW, Wik::WikJGeneral_iip, C=1)
+    @unpack WikJ, WikJ2, WikJ3, m_seq, vec_ζ, vec_η, Gp₁, Gp₂, Aᵢ = Wik
+
+    m      = length(dW)
+    M      = Int(m*(m-1)/2)
```
YingboMa: `div(m * (m - 1), 2)` or `m * (m - 1) ÷ 2` would be better, as @devmotion suggested.

src/iterated_integrals.jl
```diff
@@ -0,0 +1,345 @@
...
+
+    sum_dW² = dW' * dW #zero(eltype(dW))
```
YingboMa: I am not too familiar with `StochasticDiffEq`, but if `dW` could be a row vector, then `dW' * dW` shouldn't be used.

```julia
julia> using LinearAlgebra

julia> dW = rand(5)';

julia> dW'dW
5×5 Array{Float64,2}:
 0.240635   0.096258   0.0756145  0.172557   0.382082
 0.096258   0.0385048  0.030247   0.0690255  0.152839
 0.0756145  0.030247   0.0237603  0.0542223  0.120061
 0.172557   0.0690255  0.0542223  0.123739   0.273987
 0.382082   0.152839   0.120061   0.273987   0.606672

julia> dot(dW, dW)
1.0333109944478744
```

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,127 @@ end
...
+  if integrator.opts.adaptive
+    du₂ = integrator.f(K,p,t+dt)
+    if typeof(dW) <: Number || is_diagonal_noise(integrator.sol.prob)
+      tmp  = dt*(du₂ -du₁)/2
+      En = W.dW.^3 .* ((du₂-L)/(integrator.sqdt)).^2 / 6
+      tmp = calculate_residuals(tmp, En, uprev, u, integrator.opts.abstol,
+            integrator.opts.reltol, integrator.opts.delta, integrator.opts.internalnorm, t)
+    else
+      En = integrator.opts.internalnorm(W.dW,t)^3*ggprime_norm^2 / 6
+      tmp = @.. integrator.opts.internalnorm(dt*(du₂ - du₁)/2,t)
```
YingboMa: This doesn't look right. `@.. integrator.opts.internalnorm(...)` applies the norm to each element of the array. Do you mean

```julia
tmp = integrator.opts.internalnorm((@.. dt*(du₂ - du₁)/2),t)
```
?

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,127 @@ end
...
+      tmp = @.. calculate_residuals(tmp, En, uprev, u, integrator.opts.abstol,
```
YingboMa: Same here.

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,127 @@ end
...
+    K = @.. uprev + dt*du₁
+    utilde = (alg_interpretation(integrator.alg) == :Ito ? K : uprev) + L*integrator.sqdt
+    ggprime = (integrator.g(utilde,p,t) .- L) ./ (integrator.sqdt)
+    mil_correction = ggprime .* WikJ
+    u = K + L .* dW + mil_correction
+  else
+    for i in 1:length(dW)
+      K = uprev + dt*du₁ + sqdt*@view(L[:,i])
+      gtmp = integrator.g(K,p,t)
+      ggprime = (gtmp - L)/sqdt
```
YingboMa:
```suggestion
      ggprime = @.. (gtmp - L)/sqdt
```

src/perform_step/low_order.jl
ChrisRackauckas: it should be the latter

src/perform_step/low_order.jl
```diff
@@ -229,3 +229,127 @@ end
...
+      @.. u = K + tmp + mil_correction
+    else
+      @.. u = uprev + dt*du₁ + tmp + mil_correction
+    end
+  end
+
+  if integrator.opts.adaptive
+      En = integrator.opts.internalnorm(W.dW,t)^3*ggprime_norm^2 / 6
+      integrator.f(du₂,K,p,t+dt)
+      @.. tmp = integrator.opts.internalnorm(integrator.opts.delta * dt * (du₂ - du₁) / 2,t) + En
```
YingboMa: Same here.

src/iterated_integrals.jl
ChrisRackauckas: it shouldn't ever be a row vector, but dot is probably better anyways

src/perform_step/low_order.jl
deeepeshthakur: IIRC this is the backend definition of calculate_residuals.

src/perform_step/low_order.jl
deeepeshthakur: This error estimate is taken from `RKMil_Commute` and `RKMil` So I think it should work here too.

src/perform_step/low_order.jl
deeepeshthakur: Nevermind.

src/perform_step/low_order.jl
deeepeshthakur: And also this will change the type of `tmp`.

deeepeshthakur:
Added the suggestions except the last one.

ChrisRackauckas:
This still needs a real test for non-commutative order 1 convergence with at least a simple Milstein, but the iterated integrals are at least correct enough that I am going to merge it so it can at least be tracked. We should get that test working ASAP.

PR #218, opened 2019-08-05
error calculation in RKMil_Commute

### LINKED ISSUE #214: RKMilCommute Adaptive error calculation
@ChrisRackauckas I think `K` is not defined in this line

https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/38f408e466e77328c48b32979733ed1711242524/src/perform_step/low_order.jl#L223

deeepeshthakur:
@ChrisRackauckas 
This should fix #214

### DIFF
--- a/src/perform_step/low_order.jl
+++ b/src/perform_step/low_order.jl
@@ -204,8 +204,9 @@ end
   integrator.f(du1,uprev,p,t)
   integrator.g(L,uprev,p,t)
 
+  @.. K = uprev + dt*du1
   for j = 1:length(dW)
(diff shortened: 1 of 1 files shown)


deeepeshthakur:
@ChrisRackauckas 
We are getting convergence errors in `SKenCarp`. One of them if due undefined W
https://travis-ci.org/JuliaDiffEq/StochasticDiffEq.jl/jobs/568049932#L781

I think this is because we have
```
@unpack uf = nlsolver
```
https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/master/src/perform_step/kencarp.jl#L4

and I think 
```
@unpack uf, W = nlsolver 
```
might fix this.

PR #223, opened 2019-08-12
[WIP] SKenCarp convergence fix

kanav99:
Trying out fixes for convergence failures. Reverting back to 2\gamma is fixing the issue locally.


ChrisRackauckas:
oops yes it should be 2gamma.

PR #225, opened 2019-08-17
fix some adaptivity issues

ChrisRackauckas:


ChrisRackauckas:
@deeepeshthakur could you take this on?

PR #226, opened 2019-08-23
Upper bound dependencies

devmotion:
Same as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/891.


devmotion:
CI tests of StochasticDiffEq have been failing for some time due to errors in the stiffness detection and convergence tests of SOSRA (see, e.g, https://github.com/JuliaDiffEq/StochasticDiffEq.jl/pull/224).

I would really like them to be fixed, especially since we probably want to release a new version of StochasticDiffEq soon. @YingboMa, can you have a look at the failing stiffness detection tests? It seems the algorithm is not switching to stiff solvers anymore. Could this be caused by some recent updates of the stiffness detection and should we just adjust the test?

src/composite_algs.jl
```diff
@@ -10,17 +10,17 @@ mutable struct AutoSwitch{nAlg,sAlg,tolType,T}
   dtfac::T
   stiffalgfirst::Bool
 end
-AutoSwitch(nonstiffalg::nAlg, stiffalg::sAlg;
+AutoSwitch(nonstiffalg, stiffalg;
            maxstiffstep=10, maxnonstiffstep=3,
-           nonstifftol::T=9//10, stifftol::T=9//10,
-           dtfac=2.0, stiffalgfirst=false) where {nAlg,sAlg,T} =
+           nonstifftol=9//10, stifftol=9//10,
+           dtfac=2.0, stiffalgfirst=false) =
```
devmotion: Seems a bit strange that the default for `dtfac` is 2.0 rather than 2 (i's the same in OrdinaryDiffEq). Is there a reason for not taking an integer as default value?

YingboMa: Yeah, I should change it

devmotion:
Great, so now only the convergence failure is left :smiley:

PR #227, opened 2019-08-24
SROCKEM bug fix

deeepeshthakur:
@ChrisRackauckas @KirillZubov 
```
using StochasticDiffEq

d = 100
tspan = (0.0f0,2.0f0)
dt = 0.5
# m = 100
ts = tspan[1]:dt:tspan[2]

function F(h, p, t)
    zeros(d+1)
end
function G(h, p, t)
    zeros(d+1,d)
end

init_cond = zeros(d+1)
noise = zeros(d+1,d)
prob = SDEProblem(F, G, init_cond, tspan, noise_rate_prototype=noise)
println(prob)
predict_ans = solve(prob, SROCKEM(), dt=dt,saveat=ts)[end] #SKSROCK()
println(predict_ans)
```
Fixes the error given by the above code.


src/perform_step/SROCK_perform_step.jl
```diff
@@ -524,7 +524,7 @@ end
       for i in 1:length(W.dW)
         WikJ = W.dW[i]
         WikRange = 1//2 .* (W.dW .* WikJ - (1:length(W.dW) .== i) .* dt)
-        uᵢ₋₂ = Gₛ .* WikRange
+        uᵢ₋₂ = Gₛ*WikRange
```
ChrisRackauckas: .*?

ChrisRackauckas: nevermind, I see

PR #235, opened 2019-09-19
also test on 1.3

KristofferC:


ChrisRackauckas:
This can get expensive. What's the context on this?

KristofferC:
It failed on PkgEval for 1.3 so want to see if it is the same here.

KristofferC:
https://gist.github.com/KristofferC/47577ec5ad3b4418fd98f20bc3d91957#file-stochasticdiffeq-log

Seems to be something in DynamicPolynomials.

ChrisRackauckas:
>Might make sense to bump that to 1.2?

Yeah, that would make sense.

KristofferC:
Here is a PR that fixes this for another package https://github.com/wormell/Poltergeist.jl/pull/8. Perhaps the same could be applied to DynamicPolynomials.

KristofferC:
From that log: ` Installed DynamicPolynomials ────── v0.1.4`. This is a very old version of DynamicPolynomials. Wonder why that gets installed.

KristofferC:
https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl/pull/313 maybe...

devmotion:
Actually we had a similar (the same?) issue before in DiffEqBase: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/322#issuecomment-521553904 Since HomotopyContinuation did not support Parameters 0.11.0, we ended up with this very old version of DynamicPolynomials. The issue was solved by updating HomotopyContinuation's compatibility with Parameters IIRC

KristofferC:
Ok, so there is some old version that doesn't declare any compat with Parameters at all that the resolver finds. That's one of the "dangers" when starting to add upper bounds but there are old versions with no upper bounds. Anyway, https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl/pull/313 should fix it then.

PR #241, opened 2019-11-15
Update StaticArrays compatibility

devmotion:


devmotion:
This PR needs a new release of OrdinaryDiffEq first (StochasticDiffEq pulls in StaticArrays 0.12.1, which then makes it impossible for Pkg to find a compatible version of OrdinaryDiffEq).

PR #244, opened 2019-11-18
Make SROCKs auto-differentiable by Tracker

YingboMa:


ChrisRackauckas:
The test failures are real.

PR #254, opened 2019-12-19
-fix non-diagonal noise for LambaEM and RKMiL_General

KirillZubov:
-fix non-diagonal noise for LambaEM and RKMiL_General
-support RKMiL_General with Float32

### DIFF
--- a/src/iterated_integrals.jl
+++ b/src/iterated_integrals.jl
@@ -186,10 +186,11 @@ function get_iterated_I!(dt, dW, Wik::WikJGeneral_oop, C=1)
     sum_dW² = dot(dW,dW)
 
     WikJ = dW*dW'
-    Gp₁ = randn(M)
-    α = sqrt(1 + sum_dW²/dt)
-    Gp₂ = Gp₁/(sqrt(2)*(1+α)*dt)
-
+    eltWikJ = eltype(WikJ)
+    WikJ2 =zeros(eltWikJ, size(WikJ))
+    Gp₁ = randn(eltype(WikJ),M)
+    α = eltype(WikJ)(sqrt(1 + sum_dW²/dt))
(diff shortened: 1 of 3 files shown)


src/iterated_integrals.jl
```diff
@@ -186,10 +186,11 @@ function get_iterated_I!(dt, dW, Wik::WikJGeneral_oop, C=1)
...
 
     WikJ = dW*dW'
-    Gp₁ = randn(M)
-    α = sqrt(1 + sum_dW²/dt)
-    Gp₂ = Gp₁/(sqrt(2)*(1+α)*dt)
-
+    eltWikJ = eltype(WikJ)
+    WikJ2 =zeros(eltWikJ, size(WikJ))
+    Gp₁ = randn(eltype(WikJ),M)
+    α = eltype(WikJ)(sqrt(1 + sum_dW²/dt))
```
ChrisRackauckas: use convert instead of implicit number constructor

src/perform_step/low_order.jl
```diff
@@ -249,6 +249,8 @@ end
   du₁ = integrator.f(uprev,p,t)
   L = integrator.g(uprev,p,t)
   mil_correction = zero(u)
+  ggprime_norm = zero(eltype(u)) #0
+  # sqdt = integrator.sqdt
```
ChrisRackauckas: what's the reason here?

src/iterated_integrals.jl
KirillZubov: ok

src/perform_step/low_order.jl
ChrisRackauckas: no, the `sqdt=integrator.sqdt` getting commented, and then using the integrator in the other lines? Seems like an odd change.

Issue #182, opened 2019-06-23
Weak Order 2.0 SRK methods


ChrisRackauckas:
- [x] Ito DRI1 https://www.sciencedirect.com/science/article/pii/S0168927408000482 or https://arxiv.org/abs/1303.5103
- [x] Older ito: https://www.jstor.org/stable/27862707?seq=1
  - [x] RI1
  - [x] RI3
  - [x] RI5
  - [x] RI6
- [x] PL1WM from Kloeden and Platen
- [x] NON https://www.sciencedirect.com/science/article/pii/S0377042706003906
- [x] NON2 https://www.sciencedirect.com/science/article/pii/S0377042711002159
- [x] COM https://www.sciencedirect.com/science/article/pii/S0377042706001737?via%3Dihub
- [x] https://arxiv.org/pdf/1303.4510
  - [x] RDI1WM
  - [x] RDI2WM
  - [x] RDI3WM 
  - [x] RDI4WM
- [x] SIE / SME https://epubs.siam.org/doi/pdf/10.1137/S1064827501387814
  - [x] SIE-A
  - [x] SME-A
  - [x] SIE-B
  - [x] SME-B
- [x] Stratanovich: https://link.springer.com/article/10.1007/s10543-007-0130-3
- [x] Adaptive: https://onlinelibrary.wiley.com/doi/abs/10.1002/pamm.200410005

ChrisRackauckas:
Note that for implementing the adaptive schemes, all that needs to be done is the addition of the error estimator: https://github.com/SciML/StochasticDiffEq.jl/blob/master/src/perform_step/sri.jl#L111-L118 . If you notice the implementation of EnsembleGPUArray and EnsembleCPUArray will then automatically use the normed pooled error, so that recovers the scheme of Rossler (the Brownian bridge is handled automatically by the machinery of the package.

frankschae:
Probably, we should also add 

`COM`:
https://www.sciencedirect.com/science/article/pii/S0377042706001737?via%3Dihub
[ simplified version applicable in case of commutative noise ] 

and 

`NON2`:
https://www.sciencedirect.com/science/article/pii/S0377042711002159
[ modified version with lower computational costs ]

ChrisRackauckas:
GG @frankschae

PR #269, opened 2020-02-27
[WIP] Weak Order 2.0 SRK methods #182

frankschae:
I added the constant cache part of the DRI1 method from the newer Ito paper.
I don't obtain the second order convergence on the stochastic part. 
The deterministic part converges with order 3 as expected. The associated test is implemented in ode_convergence_regression.jl


ChrisRackauckas:
```julia
using Distributions
quantile(Normal(),1/6) # -0.9674215661017014
```

Declare this a constant and use the following:

```julia
dW_scaled = W.dW / sqrt(dt)
if _dW < -NORMAL_ONESIX_QUANTILE
  _dW = -sqrt(3dt)
elseif _dW > NORMAL_ONESIX_QUANTILE
  _dW = sqrt(3dt)
else
  _dW = zero(dt)
end
```

but in a `map` over the values.

ChrisRackauckas:
Choose +- integrator.sqdt based on `sign(W.dZ)` and make this a process with two Brownians.

src/perform_step/srk_weak.jl
```diff
@@ -0,0 +1,164 @@
...
+    H12 = [uprev + a121*k1*dt .+ b121*g1[k]*integrator.sqdt for k=1:m]
+  else
+    H12 = [uprev + a121*k1*dt + b121*g1[:,k]*integrator.sqdt for k=1:m]
+  end
+
+  # # H_i^(k), stage 3
+  if typeof(W.dW) <: Number
+    H13 = uprev + a131*k1*dt + b131*g1*integrator.sqdt
+  elseif is_diagonal_noise(integrator.sol.prob)
+    H13 = [uprev + a131*k1*dt .+ b131*g1[k]*integrator.sqdt for k=1:m]
```
ChrisRackauckas: these comprehensions should be broadcasts.

src/alg_utils.jl
```diff
@@ -57,6 +57,9 @@ alg_order(alg::SRA2) = 2//1
 alg_order(alg::SRA3) = 2//1
 alg_order(alg::SOSRA) = 2//1
 alg_order(alg::SOSRA2) = 2//1
+
+alg_order(alg::DRI1) = 3//2 # deterministic // weak
```
ChrisRackauckas: should be 1?

ChrisRackauckas:
Check the order conditions in terms of the beta parts. My guess is that one of those are actually incorrect if it's not converging below 0.01.

frankschae:
[EM-like.pdf](https://github.com/JuliaDiffEq/StochasticDiffEq.jl/files/4268033/EM-like.pdf)
The EM test look better now with the new sampling you proposed. I have not checked the conditions for the betas yet but what I found a bit odd is that also for the RI1 tableau (from the old ito paper) I don't find convergence.

frankschae:
> [EM-like.pdf](https://github.com/JuliaDiffEq/StochasticDiffEq.jl/files/4268033/EM-like.pdf)
> The EM test look better now with the new sampling you proposed. I have not checked the conditions for the betas yet but what I found a bit odd is that also for the RI1 tableau (from the old ito paper) I don't find convergence.

ChrisRackauckas:
Note that the other method is called the Simplified Euler-Maruyama scheme (SimplifiedEM()), and would be good to add as well. It should weakly converge with order 1.

http://www.diva-portal.se/smash/get/diva2:861891/FULLTEXT01.pdf

and it originates from KPS.

src/caches/srk_weak_caches.jl
```diff
@@ -0,0 +1,178 @@
...
+
+  #beta41::T
+  beta42::T
+  beta43::T
+
+  #quantile(Normal(),1/6)
+  NORMAL_ONESIX_QUANTILE::T
+end
+
+function DRI1ConstantCache(T::Type, T2::Type)
```
ChrisRackauckas, line 68:
```suggestion
function DRI1ConstantCache(::Type{T}, ::Type{T2}) where {T1,T2}
```

ChrisRackauckas:
can you share the current long convergence plots here?

YingboMa:
My inference fix is here https://github.com/SciML/StochasticDiffEq.jl/commit/ba3fa77f928554c9a79b2abd4cfdcdddcf87a645

frankschae:
[DRI1-CPU-final-500000.0.pdf](https://github.com/SciML/StochasticDiffEq.jl/files/4423700/DRI1-CPU-final-500000.0.pdf)
[RI1-CPU-final-500000.0.pdf](https://github.com/SciML/StochasticDiffEq.jl/files/4423702/RI1-CPU-final-500000.0.pdf)

frankschae:
The plots above  for DRI1 and RI1 are generated using the new file "srk_weak_final.jl"

ChrisRackauckas:
That new test script isn't being ran? How long does it take to run?

frankschae:
Only a few minutes..
The @ times of the script print:
28.555001 seconds (778.67 M allocations: 80.634 GiB, 34.83% gc time)
29.302624 seconds (841.61 M allocations: 81.294 GiB, 36.58% gc time)

ChrisRackauckas:
Awesome, add it to the runtests and let's mark this as not WIP but done. We can continue to improve it later.

ChrisRackauckas:
Awesome. Please open an issue on the inference bug so we can have the compiler team work on that. Let's continue refining the tests next, and add an in-place version.

PR #270, opened 2020-02-29
SimplifiedEM() scheme from discussion in #182

frankschae:
Quite a large number of trajectories is needed in the weak tests to get a "pass". I haven't found the strong convergence order .. is it right to enter 0.5 for alg_order?


src/perform_step/low_order.jl
```diff
@@ -104,6 +104,44 @@ end
...
+  @unpack rtmp1,rtmp2 = cache
+  @unpack t,dt,uprev,u,W,p = integrator
+
+  integrator.f(rtmp1,uprev,p,t)
+
+  @.. u = uprev + dt * rtmp1
+
+  integrator.g(rtmp2,uprev,p,t)
+
+  _dW = map(x -> sign(x) > 0.0 ? integrator.sqdt : -integrator.sqdt, W.dW)
```
ChrisRackauckas: the cache version should be non-allocating, so make a `_dW` cache and then fill it with a `map!`. Pull the anonymous function outside of the function to be some generic function so it precompiles.

ChrisRackauckas:
>is it right to enter 0.5 for alg_order?

Yes

>Quite a large number of trajectories is needed in the weak tests to get a "pass".

Yes, that is expected. It's an average of averages, and the average will converge as sqrt(N), so you'll need N^2 simulations to get the average "correct" enough to be seen in the convergence simulation. This may be the cause of your other issue. You may want to make sure it's using EnsembleGPUArray() and then check your other method if you have the compute resources for that.

Essentially here's what's going on. Averages can only be correct to `C/sqrt(N)` by the central limit theorem, and so even if "`dt` is zero", you will only be C/sqrt(N) away from the true average simply because the expectation is a limit as N->infinity. What this means is that there will be a floor where the dominant error is no longer discretization error but is now sampling error.

[DRI1.pdf](https://github.com/JuliaDiffEq/StochasticDiffEq.jl/files/4270610/DRI1.pdf)

With this understanding, when I look at your convergence plot for DRI1, I think that's what I'm seeing: second order convergence that then saturates at the value of the sampling error. The way to fix this would be to increase dt (to get more points that are "easy") and then also increase the number of samples. If this is done then you should get a good solid line.

src/perform_step/low_order.jl
frankschae: Okay! I also already thought about it but I wasn't sure. Should I put the function in a separate weak_utils.jl file? I guess the two-point and three-point distributed random variables will be used in several places for the next solvers

ChrisRackauckas: >Okay! I also already thought about it but I wasn't sure. Should I put the function in a separate weak_utils.jl file? I guess the two-point and three-point distributed random variables will be used in several places for the next solvers

Yup, makes complete sense.

PR #271, opened 2020-02-29
enable weak convergence on CI

ChrisRackauckas:
It's costly, but now with the latest work we will definitely want to be tracking weak convergence orders, so let's get this going. 

I split the tests and did some renaming. Hopefully they all fit in the 50 minutes for one Travis job, otherwise we can split them into separate Travis jobs. If necessary, we can make this use the GPU CI.


frankschae:
it looks to me that it fails because it hits the "no log output for more than 10 minutes" requirement. Then splitting the individual SDE tests to independent groups should not solve it (and for prob_sde_2Dlinear 10^4-10^5 trajectories take a while.. so even splitting on the level of single solvers might not work).  Should we then go for GPU CI? (I guess artificially producing some intermediate log output is not what we want)

ChrisRackauckas:
That CI isn't going to work either...

.gitlab-ci.yml
```diff
@@ -0,0 +1,52 @@
...
+  extends: .root
+  stage: build
+  script:
+    - curl https://julialang-s3.julialang.org/bin/linux/x64/$(echo $JULIA_VERSION | cut -d '.' -f 1,2)/julia-${JULIA_VERSION}-linux-x86_64.tar.gz -o julia.tar.gz
+    - unp julia.tar.gz
+    - export PATH="$(pwd)/julia-${JULIA_VERSION}/bin:$PATH"
+    - mkdir $JULIA_DEPOT_PATH # Pkg.jl#325
+    - julia -e "using InteractiveUtils;
+                versioninfo()"
+    - julia --project -e "using Pkg; Pkg.build(); using OrdinaryDiffEq;"
```
frankschae, line 29: - julia --project -e "using Pkg; Pkg.build(); using StochasticDiffEq;"

test/weak_convergence/additive_weak.jl
```diff
@@ -0,0 +1,181 @@
+using StochasticDiffEq, DiffEqDevTools, Test
+using Random
+using DiffEqProblemLibrary.SDEProblemLibrary: importsdeproblems
+importsdeproblems()
+using DiffEqProblemLibrary.SDEProblemLibrary: prob_sde_linear, prob_sde_2Dlinear, prob_sde_additive
```
frankschae, line 5: using DiffEqProblemLibrary.SDEProblemLibrary: prob_sde_additive

test/runtests.jl
```diff
@@ -51,17 +51,22 @@ const is_APPVEYOR = Sys.iswindows() && haskey(ENV,"APPVEYOR")
...
-    @time @safetestset "Roessler weak SRK Tests" begin include("srk_weak_final.jl") end
     @time @safetestset "Rossler Order Tests" begin include("sde/sde_rosslerorder_tests.jl") end
     @time @safetestset "ODE Convergence Regression Tests" begin include("ode_convergence_regression.jl") end
     @time @safetestset "Additive SDE Tests" begin include("sde/sde_additive_tests.jl") end
     @time @safetestset "Split Tests" begin include("split_tests.jl") end
     @time @safetestset "Stratonovich Convergence Tests" begin include("stratonovich_convergence_tests.jl") end
   end
+
+  if !is_APPVEYOR && (GROUP == "All" || GROUP == "WeakConvergence")
+    @time @safetestset "Roessler weak SRK Tests" begin include("srk_weak_final.jl") end
```
frankschae, line 67: @time @safetestset "Roessler weak SRK Tests" begin include("weak_convergence/srk_weak_final.jl") end

ChrisRackauckas:
@frankschae you said the really long ones can be changed to run with less trajectories? I'll merge it like this, and then a follow up PR can handle that.

PR #273, opened 2020-03-02
non-allocating two-point random variable

frankschae:
I separated the generation of the random variable to a new file weak_utils.jl that contains a in-place and out-of-place version. I had the feeling that the ternary operator performs (slightly) better than elseif.


ChrisRackauckas:
Interesting that the ternary is faster, wouldn't've expected that here.

PR #274, opened 2020-03-05
Allow separate jac_prototype and sparsity

mauro3:
Make it compatible with https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/449.

I really don't know this code, so I couldn't judge whether `jac_prototype` could/should be replaced with `sparsity`.  See https://github.com/JuliaDiffEq/StochasticDiffEq.jl/search?q=jac_prototype&type=Code.

Also note that the only [file](https://github.com/JuliaDiffEq/StochasticDiffEq.jl/blob/03d695e07aeffeb525b2cc84892e94f0292bbee2/test/sparsediff_tests.jl) which kinda tests this is not run through the test suite and does indeed not pass (at least with this PR).

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -14,6 +14,7 @@ const is_APPVEYOR = Sys.iswindows() && haskey(ENV,"APPVEYOR")
     @time @safetestset "Static Array Tests" begin include("static_array_tests.jl") end
     @time @safetestset "Noise Type Tests" begin include("noise_type_test.jl") end
     @time @safetestset "Mass matrix tests" begin include("mass_matrix_tests.jl") end
+    #@time @safetestset "Sparse Jacobian tests" begin include("sparsediff_tests.jl") end
     @time @safetestset "Outofplace Arrays Tests" begin include("outofplace_arrays.jl") end
     @time @safetestset "tdir Tests" begin include("tdir_tests.jl") end
     @time @safetestset "tstops Tests" begin include("tstops_tests.jl") end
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -14,6 +14,7 @@ const is_APPVEYOR = Sys.iswindows() && haskey(ENV,"APPVEYOR")
     @time @safetestset "Static Array Tests" begin include("static_array_tests.jl") end
     @time @safetestset "Noise Type Tests" begin include("noise_type_test.jl") end
     @time @safetestset "Mass matrix tests" begin include("mass_matrix_tests.jl") end
+    #@time @safetestset "Sparse Jacobian tests" begin include("sparsediff_tests.jl") end
```
ChrisRackauckas: why did this get commented out?

mauro3: No, it wasn't there at all.  I decided to add it as a commented out line so it is not forgotten.  I didn't check whether those tests actually work on master.  They do fail on this PR though.

mauro3: I now ran that file on master as well and it gives the same failure.  On L49 the error is too big:
```
julia> maximum(abs.(sol_sp.u[end].-sol.u[end]))
0.0879739350640969
```
and on the next line the lengths are different
```
julia> length(sol_sp.t), length(sol.t)
(28, 34)
```
Similarly for the other two tests.

ChrisRackauckas: Oh I see... that test is bad. It's using a different Brownian process between the two, so of course the solution is stochastic haha. That needs some work.

mauro3: I thought there should be some random seed needing to be set.  So this is ready to be merged, I think.

ChrisRackauckas: add seed=1 to `solve` and see?

If it doesn't work, I can take it from here, but I would like to know the result.

mauro3: No, that doesn't help, at all.  On master I still get the identical numbers I reported above.

PR #282, opened 2020-04-07
better type inference in srk weak

ChrisRackauckas:


ChrisRackauckas:
@frankschae looks like weak tests fail: https://gitlab.com/juliadiffeq/StochasticDiffEq-jl/-/jobs/501261403#L89

frankschae:
Right.. I check what is going on here .. Do the fixed random seeds 
```
using Random
Random.seed!(100)
```
lead cross-platform to the same random sequences for the random variables?

frankschae:
Oh I see that Random.seed! doesn't fix the final outcome for the slope. In many cases the tests pass but sometimes they don't.

PR #289, opened 2020-05-11
iip for RI1 with tests

frankschae:
I added `alg_cache()` and further tests such that now also `RI1()` works for the inplace problem.
(I realized that` @test m-2 < 0.3 `from before didn't make sense because this would be true also for m < 1.7 but I wanted it to be true for m> 2.3 .. so I added the minus signs.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
   - GROUP=AlgConvergence
   - GROUP=AlgConvergence2
   - GROUP=AlgConvergence3
+  - GROUP=WeakConvergence
 #matrix:
 #  allow_failures:
 #    - julia: nightly
--- a/src/caches/srk_weak_caches.jl
+++ b/src/caches/srk_weak_caches.jl
@@ -250,3 +250,44 @@ function alg_cache(alg::DRI1,prob,u,ΔW,ΔZ,p,rate_prototype,
 
   DRI1Cache(u,uprev,_dW,_dZ,chi1,Ihat2,tab,g1,g2,g3,k1,k2,k3,H02,H03,H12,H13,H22,H23,tmp1,tmpg)
 end
+
+
+function alg_cache(alg::RI1,prob,u,ΔW,ΔZ,p,rate_prototype,
+                   noise_rate_prototype,uEltypeNoUnits,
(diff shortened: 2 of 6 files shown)


frankschae:
Is there an easy way to (slightly) increase the 10 minutes timeout?  
With other processes running in the background, it takes 
```
712.479824 seconds (7.16 G allocations: 733.392 GiB, 44.16% gc time)
DRI1:2.4826563100491574
670.594097 seconds (7.86 G allocations: 744.664 GiB, 35.23% gc time)
RI1:2.9407305307490668
```
on my machine to start

ChrisRackauckas:
Not on Travis, and Travis is probably slower than your machine. Try to relax it a bit?

frankschae:
This time it got aborted at the computation for the third test which should however not take ten minutes, given that the first two tests passed.

Should I maybe rewrite the list comprehensions as loops with `println` for each `dt` or should I try to decrease further the number of trajectories and/or the tolerances for the tests?

These are the timings on my machine..

```
julia srk_weak_final.jl 
start with scalar
390.342434 seconds (4.31 G allocations: 443.118 GiB, 38.90% gc time)
DRI1:2.27094894385398
389.970990 seconds (4.73 G allocations: 449.621 GiB, 32.01% gc time)
RI1:2.0231138501642105
305.206892 seconds (2.19 G allocations: 224.077 GiB, 19.21% gc time)
DRI1:2.8151946151968485
343.055873 seconds (2.61 G allocations: 230.438 GiB, 19.31% gc time)
RI1:1.8218004380858268
356.693658 seconds (1.54 G allocations: 145.723 GiB, 9.46% gc time)
DRI1:1.84272672406444
397.042005 seconds (1.89 G allocations: 157.582 GiB, 9.60% gc time)
RI1:1.5246185931766558
 97.152105 seconds (722.58 M allocations: 80.761 GiB, 14.19% gc time)
DRI1:2.2407510453590698
109.571199 seconds (864.97 M allocations: 83.065 GiB, 14.26% gc time)
RI1:2.376887074688166
```

ChrisRackauckas:
Travis is a lot slower than other computers because it gets oversubscribed. You may want to split these tests in two and keep tweaking them until it starts working out. This could be a long process of waiting, so just make sure you're doing something else in parallel.

PR #292, opened 2020-05-14
utilize OrdinaryDiffEq's implicit handling structures

ChrisRackauckas:
Unifies the packages in term of handling implicit equations


devmotion:
Nice, StochasticDiffEq (and StochasticDelayDiffEq) should really get all the goodies that are already used by OrdinaryDiffEq.

ChrisRackauckas:
StochasticDelayDiffEq really needs an update whenever someone has time. I don't know if it really works still. We just need to get it updated one time and get proper downstream tests like DelayDiffEq has and we can keep it maintained, but we never really went far enough the first time to get it going.

devmotion:
> StochasticDelayDiffEq really needs an update whenever someone has time. I don't know if it really works still. We just need to get it updated one time and get proper downstream tests like DelayDiffEq has and we can keep it maintained, but we never really went far enough the first time to get it going.

Yep, it's a bit unfortunate. The tests pass but the amount of tests is quite limited (57% coverage apparently so far). I was thinking about having a closer look at it in June, if nobody beats me to it :smiley:

src/alg_utils.jl
```diff
@@ -195,3 +194,16 @@ end
 
 issplit(::StochasticDiffEqAlgorithm) = false
 issplit(::SplitSDEAlgorithms) = true
+
+function OrdinaryDiffEq.unwrap_alg(integrator::SDEIntegrator, is_stiff)
+  alg = integrator.alg
+  iscomp = typeof(alg) <: StochasticCompositeAlgorithm
+  if !iscomp
```
devmotion, line 201: Maybe use
```suggestion
  if !is_composite(alg)
```
?

src/alg_utils.jl
```diff
@@ -195,3 +194,16 @@ end
...
+    return alg
+  elseif typeof(alg.choice_function) <: AutoSwitch
+    num = is_stiff ? 2 : 1
+    return alg.algs[num]
+  else
+    return alg.algs[integrator.cache.current]
+  end
+end
```
devmotion, line 209: Just noticed that `unwrap_alg` is actually implemented as suggested in the comment above. So maybe just
```suggestion
function OrdinaryDiffEq.unwrap_alg(integrator::SDEIntegrator, is_stiff)
  return unwrap_alg(integerator, is_stiff)
end
```
? Or can `StochasticDiffEq.unwrap_alg` be removed?

devmotion:
As discussed before at some point, maybe it would make sense to move these methods to a separate package in a next step (but maybe not DiffEqBase) that both StochasticDiffEq and OrdinaryDiffEq could use, without having to depend on eachother?

src/alg_utils.jl
ChrisRackauckas: it's a typing issue and I think it's due to how DelayDiffEq is calling into it as well. So at least for now we need it, and we can continue to clean this up.

PR #299, opened 2020-05-16
post leap adaptive tau leaping

ChrisRackauckas:
Implemented is post-leap adaptive tau leaping. The Anderson paper has a typo because the rates should be scaled by dt and this is corrected in the implementation. The scheme is adapted into a proportional error control that seems to work quite well.

One thing to note is that the method itself introduces a funny error that's not mentioned in the paper. Unlike Brownian motion based methods, the noise process itself has an error if the rates are off, i.e. if dt is large. This means that Anderson's method is not exact, which is different from the SDE adaptive method. Although it seems exact, when there are multiple jumps the rates will change, and if the rates change you get the wrong number of jumps in the interval. You will still interpolate it well, but this error persists even with the Binomial interpolation. But given the leap condition is checked exactly, you still get the good properties.

Basically it's highlighting that the leap condition is kind of funny because it's not "exact" and has an epsilon parameter in there, so it does weird thing when epsilon is too close to one. Since this is a relative tolerance, we make it `reltol`. `reltol=5e-2` as high as possible because we internally do `50integrator.opts.reltol` for rescaling to a good default, which means `reltol=5e-2` has epsilon = 1 which is the maximum value that could possibly be allowed. So you can still go negative even if the leap condition is satisfied, just because this leap condition is ill-defined if epsilon is not << 1.

When reltol is kept sane, indeed the leap condition is held and the algorithm works. It's scaled so default tolerances are large but still in the safe zone. This will need more tweaking to be perfect, but it seems pretty good.

```julia
@time sol = solve(jump_iipprob,TauLeaping())
```


ChrisRackauckas:
```julia
using StochasticDiffEq, DiffEqJump, DiffEqBase
using Test, LinearAlgebra, Plots

function regular_rate(out,u,p,t)
    out[1] = (0.1/1000.0)*u[1]*u[2]
    out[2] = 0.01u[2]
end

const dc = zeros(3, 2)
dc[1,1] = -1
dc[2,1] = 1
dc[2,2] = -1
dc[3,2] = 1

function regular_c(du,u,p,t,counts,mark)
    mul!(du,dc,counts)
end

rj = RegularJump(regular_rate,regular_c,2)
jumps = JumpSet(rj)
iip_prob = DiscreteProblem([999,1,0],(0.0,250.0))
jump_iipprob = JumpProblem(iip_prob,Direct(),rj)
sol = solve(jump_iipprob,TauLeaping())
plot(sol)
```

plot

src/solve.jl
```diff
@@ -119,8 +119,10 @@ function DiffEqBase.__init(
   tTypeNoUnits   = typeof(one(tType))
 
   if abstol === nothing
-    if uBottomEltypeNoUnits == uBottomEltype
+    if uBottomEltypeNoUnits == uBottomEltype && !(uBottomEltype <: Integer)
```
devmotion, line 122: Maybe
```suggestion
    if uBottomEltypeNoUnits === uBottomEltype && !(uBottomEltype <: Integer)
```
?

src/solve.jl
```diff
@@ -129,8 +131,10 @@ function DiffEqBase.__init(
   end
 
   if reltol === nothing
-    if uBottomEltypeNoUnits == uBottomEltype
+    if uBottomEltypeNoUnits == uBottomEltype && !(uBottomEltype <: Integer)
```
devmotion, line 134:
```suggestion
    if uBottomEltypeNoUnits === uBottomEltype && !(uBottomEltype <: Integer)
```

PR #305, opened 2020-05-18
tstop checks for reverse time propagation

frankschae:
Fixes an error which I got for SDE adjoints where propagation from (tend, tstart) with tend>tstart is used, see https://github.com/SciML/DiffEqSensitivity.jl/pull/242.


ChrisRackauckas:
Rebased. Make sure to work off of current masters.

PR #325, opened 2020-06-08
Use JuliaGPU's GitLab CI setup

christopher-dG:


ChrisRackauckas:
@frankschae we can now setup some GPU-based tests

frankschae:
Can I use `JULIA_NUM_THREADS: 4` somehow to  run each dt of the convergence simulation on a different thread, or does this simply mean that for each computation 4 threads are used?

ChrisRackauckas:
It'll run different trajectories on different threads, yes. It'll actually parallelize over the trajectories of a same dt at a time since that's a bit more efficient (they will take all the same time so static schedules will do better), but same difference. We can use this on the Gitlab runners, though we'd need to set a tag so we draw from a pool with good CPUs if we do that.

PR #331, opened 2020-06-16
[WIP] adaptive weak solver

frankschae:
The embedded scheme is chosen to have second order convergence on the drift term and first order (weak) convergence. This is achieved by choosing alpha^ such that it fulfills alpha^*e =1 and alpha^*c=1//2, i.e., the order conditions for a deterministic RK method + defining beta1 = [1,0,0] and all the other betas = 0. I am not sure if one can do much better.. I tried to keep a few betas non-zero to have a few second order conditions fulfilled but I didn't had the feeling that this really changes something.. 

[ for the methods where c02 ==0, the embedded scheme is the Simplified EM method, as those weak SRK with c02 ==0 anyway just have a second order deterministic convergence]

I still have to add some tests (like those "Linear SDE Tests" begin include("sde/sde_linear_tests.jl") end). It would be great if we could check first if the current solution goes into the right direction.

Please find attached the Figure how DRI1 choses the time steps on the Brusselator example. (time range is (1,100), x axis in the plot shows the step index, y axis shows dt)
chosen_timesteps

related issue: https://github.com/SciML/StochasticDiffEq.jl/issues/182

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -24,6 +24,7 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 SparseDiffTools = "47a9eef4-7e08-11e9-0b38-333d64bd3804"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 
 [compat]
@@ -49,6 +50,7 @@ julia = "1.2"
 [extras]
 DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
 DiffEqDevTools = "f3b72e0c-5b89-59e1-b016-84e28bfd966d"
+DiffEqGPU = "071ae1c0-96b5-11e9-1965-c90190d839ea"
 DiffEqOperators = "9fdde737-9c7f-55bf-ade8-46b3f136cc48"
 DiffEqProblemLibrary = "a077e3f3-b75c-5d7f-a0c6-6bc4c8ec64a9"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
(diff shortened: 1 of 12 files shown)


Project.toml
```diff
@@ -57,4 +59,4 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["DiffEqCallbacks", "DiffEqDevTools", "DiffEqOperators", "DiffEqProblemLibrary", "SafeTestsets", "SparseArrays", "Statistics", "Test"]
+test = ["DiffEqCallbacks", "DiffEqDevTools", "DiffEqGPU", "DiffEqOperators", "DiffEqProblemLibrary", "SafeTestsets", "SparseArrays", "Statistics", "Test"]
```
ChrisRackauckas, line 62: GPUs should get a different Project.toml that's instantiated. See DiffEqBase

src/perform_step/srk_weak.jl
```diff
@@ -152,8 +152,32 @@
...
+      # check against EM
+      uhat = uprev + k1*dt
+    end
+    if is_diagonal_noise(integrator.sol.prob)
+      uhat += g1 .* _dW
+    else
+      uhat += g1 * _dW
+    end
+
+    Eprev = Statistics.mean(uprev,dims=2)
```
ChrisRackauckas, line 174: I don't think you need to take a mean on these, since it will naturally happen as part of the norm on the resids

ChrisRackauckas: it won't actly be a mean, but it'll be an average of sorts

frankschae: oh yes.. I just checked that they are indeed quite close by printing the values for the Brusselator example. .. at first I simply checked it with `isapprox()` (which is false.. therefore I went initially for the version with the mean to be as close as possible to Roesslers description.)

Project.toml
frankschae: @ChrisRackauckas  any idea what is still missing for the GPU tests? Looks like they are not starting..

ChrisRackauckas:
Oh they don't start from forks. Let me just invite you. You have 2FA enabled?

frankschae:
These plots are generated from the last lines in sde_weak_adaptive.jl .. Somehow it doesn't look there that the estimation of the expected value gets better with smaller abstol.

non-adaptive
adaptive
adaptive-1e-2

frankschae:
Yes,  I have 2FA enabled.

PR #332, opened 2020-06-17
Adaptive srk

frankschae:
Continues PR #331 from another branch that is not from my fork. Adaptive scheme for stochastic Runge-Kutta methods due to Roessler #182 .

PR #333, opened 2020-06-17
Stratonovich SRK methods due to Roessler

frankschae:
Adds RS1 and RS2 [1]  for the weak approximation in the Stratonovich sense from issue #182 .
Tests for  diagonal and non-commuting noise are taken from [1]. 

[[1]Roessler's paper]( https://link.springer.com/content/pdf/10.1007/s10543-007-0130-3.pdf)

### DIFF
--- a/src/StochasticDiffEq.jl
+++ b/src/StochasticDiffEq.jl
@@ -127,7 +127,8 @@ module StochasticDiffEq
          SROCK1, SROCK2, SROCKEM, SKSROCK, TangXiaoSROCK2, KomBurSROCK2, SROCKC2,
          WangLi3SMil_A, WangLi3SMil_B, WangLi3SMil_C, WangLi3SMil_D, WangLi3SMil_E, WangLi3SMil_F,
          AutoSOSRI2, AutoSOSRA2,
-         DRI1, RI1, RI3, RI5, RI6, RDI1WM, RDI2WM, RDI3WM, RDI4WM
+         DRI1, RI1, RI3, RI5, RI6, RDI1WM, RDI2WM, RDI3WM, RDI4WM,
+         RS1, RS2
 
   export EulerHeun, LambaEulerHeun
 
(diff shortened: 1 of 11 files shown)


src/StochasticDiffEq.jl
```diff
@@ -151,7 +152,6 @@ module StochasticDiffEq
   export solve, init, solve!, step!
 
   #Misc Tools
-  export checkSRIOrder, checkSRAOrder,  checkRIOrder, constructSRIW1, constructSRA1, constructDRI1, constructRI1, constructRI3, constructRI5, constructRI6,
-        constructRDI1WM, constructRDI2WM, constructRDI3WM, constructRDI4WM
+  export checkSRIOrder, checkSRAOrder,  checkRIOrder, checkRSOrder, constructSRIW1, constructSRA1, constructDRI1, constructRI1, constructRI3, constructRI5, constructRI6, constructRDI1WM, constructRDI2WM, constructRDI3WM, constructRDI4WM, constructRS1, constructRS2
```
ChrisRackauckas, line 155: whatta line

frankschae: I added some line breaks 👍

frankschae:
Had to create a 4th group of weak tests -- they are too long to fit together with other tests in the 50 min Travis window. 

Once the adaptive schemes from PR #332 are merged, I'll start moving also weak tests to the GPU.  (The CI GPU tests however still don't seem to start. They freeze at pending.)

ChrisRackauckas:
https://gitlab.com/JuliaGPU/StochasticDiffEq.jl/-/jobs/600755893 they are fine, just the section is empty for them right now.

ChrisRackauckas:
We need to move to other GPU CI machines first though. I think Tim might be getting a bit frustrated with us killed machines in the OrdinaryDiffEq one haha

PR #334, opened 2020-06-23
Pl1WM scheme

frankschae:
This PR adds the PL1WM method due to Platen from issue #182  for the general noise case, i.e., Eq.~(15.1.3).


ChrisRackauckas:
You'll want to bump the DiffEqNoiseProcess lower bound to 5

frankschae:
yes, thanks! Do you think it is worth to add the additive noise case as a separate alg (PL1WMA or so)?  It is given in Kloeden & Platen as well [ Eq.(15.1.4)]

ChrisRackauckas:
Yes, additive noise cases come up a lot, and they are usually quite a bit faster.

PR #337, opened 2020-06-28
NON

frankschae:
The scalar noise tests are passing for the inplace and out-of-place version of NON from issue #182 . 
Also ODE tests are passing (4th order, basically standard RK4).
I have checked that all coefficients in the paper satisfy the order conditions. 

The diagonal noise `perform_step`  actually does in one step what I expect it to do, i.e.,  for the SDE
```
using StochasticDiffEq
using Random
u₀ = [0.1,0.1]
function f3!(du,u,p,t)
  du[1] = p[1]*u[1]
  du[2] = p[1]*u[2]
end
function g3!(du,u,p,t)
  du[1] = p[2]*u[1]
  du[2] = p[2]*u[1]
end
tspan = (0.0,1.0)
p = [299//200, 1//10]

prob = SDEProblem(f3!,g3!,u₀,tspan,p)
ensemble_prob = EnsembleProblem(prob;
        prob_func = prob_func
        )
sol = solve(ensemble_prob, NON(); dt=1., trajectories=numtraj, ensemblealg=EnsembleThreads())
# [0.469606, 0.409286]
```
So my expectation is that I misunderstand something in the stage values (Y^{j,l}) in the paper (or something else is odd..)

The convergence plot for diagonal noise is shown in the attached figure. The slope is ~ 0.8.  Other solvers in the Stratonovich sense such as `RS1` or `EulerHeun`  are passing the test. Therefore, the example should be ok. 

NON_diagonal_noise

### DIFF
--- a/test/weak_convergence/weak_strat.jl
+++ b/test/weak_convergence/weak_strat.jl
@@ -193,7 +193,7 @@ _solutions = @time generate_weak_solutions(ensemble_prob, NON(), dts,
 errors = [LinearAlgebra.norm(Statistics.mean(sol.u) .- u₀.*exp(1.0*(p[1]+0.5*p[2]^2))) for sol in _solutions]
 m = log(errors[end]/errors[1])/log(dts[end]/dts[1])
 
-using Plots; convergence_plot = plot(dts, errors, xaxis=:log, yaxis=:log)
+
 """
  Test non-commutative noise SDEs (iip)
 """
@@ -249,6 +249,19 @@ m = log(errors[end]/errors[1])/log(dts[end]/dts[1])
 
 println("RS2:", m)
 
+numtraj = Int(1e6)
+seed = 100
+Random.seed!(seed)
+seeds = rand(UInt, numtraj)
+
+_solutions = @time generate_weak_solutions(ensemble_prob, NON(), dts, numtraj, ensemblealg=EnsembleThreads())
+
+errors = [LinearAlgebra.norm(Statistics.mean(sol.u)-1//100*exp(2)) for sol in _solutions]
+m = log(errors[end]/errors[1])/log(dts[end]/dts[1])
+@test -(m-2) < 0.3
+
+println("NON:", m)
+
 """
  Test Diagonal noise SDEs (iip)
 """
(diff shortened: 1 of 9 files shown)


test/weak_convergence/weak_strat.jl
```diff
@@ -108,6 +110,27 @@ m = log(errors[end]/errors[1])/log(dts[end]/dts[1])
...
+        output_func = (sol,i) -> (h1(sol[end]),false),
+        prob_func = prob_func
+        )
+
+
+_solutions = @time generate_weak_solutions(ensemble_prob, NON(), dts, numtraj, ensemblealg=EnsembleThreads())
+
+errors = [LinearAlgebra.norm(Statistics.mean(sol.u) .- u₀.*exp(1.0*(p[1]+0.5*p[2]^2))) for sol in _solutions]
+m = log(errors[end]/errors[1])/log(dts[end]/dts[1])
+@test -(m-2) < 0.3
```
ChrisRackauckas, line 130: these tests should really be on the absolute value

ChrisRackauckas:
chop off the highest dt there: that's before the limiting behavior kicks in

frankschae:
Without the largest dt value, the slope is ~0.94.. so looks almost like perfect first order convergence.. 

NON_diagonal_noise

test/weak_convergence/weak_strat.jl
frankschae: Isn't norm the same as the absolute value for scalars/1d arrays https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.norm ?

ChrisRackauckas: First though, is there a reason why this file doesn't use the standard convergence simulation? It's doing the same calculation. But specifically right here, this doesn't capture both directions, i.e. `abs(m-2) < 0.3`. We can probably finish up all of the weak methods first and then turn around and make this use the standard tools. It's not a big deal, but it does make maintenance easier down the road.

frankschae: IIRC the problem was that much more memory was used in the standard tools (potentially because more than the weak final error was tested?) and then it always filled up my swap. 

Yes  (by tweaking the used random seed or using more trajectories) we could use  `abs(m-2) < 0.3` ..  at the moment because of the smallest dt value, we sometimes get a slightly better convergence.

frankschae:
@ChrisRackauckas  tests here are looking good now. It is written a bit cumbersome because terms like alpha_ljjl are also  computed  in the case of diagonal noise  or commutative non-diagonal noise (They should vanish in principle but earlier tests without those computations failed..).  I think the best way would be to add the `COM` scheme which is the same as NON but with a restricted Butcher Table -- such that if someone knows that he doesn't need a solver for non-commutative noise,  he probably anyways wants to choose COM instead of NON.

ChrisRackauckas:
Yeah, it does look a little bit wasteful, but I guess that's because COM fixes that so this is fine.

PR #341, opened 2020-07-18
Fix logging for `progress=true` by wrapping log-level `-1` in `LogLevel(-1)`

jleugeri:
This change was necessary to get  `solve(SDEProblem(...),args...;kwargs..., progress=true)` to work. 
Before, this threw `MethodError: no method matching isless(::Int64, ::Base.CoreLogging.LogLevel)`


ChrisRackauckas:
Thanks!

PR #343, opened 2020-07-27
Com

frankschae:
This adds the `COM` scheme which is the commutative noise version of the NON scheme due to Komori (Refs are in issue #182 ).
A bit odd is that `COM` passes our non-commuting noise tests -- there is a related plot in their paper showing that one has to use  small values of dt to see a saturation. However, this would mean that we'd have to significantly increase the number of trajectories in the tests.

### DIFF
--- a/src/StochasticDiffEq.jl
+++ b/src/StochasticDiffEq.jl
@@ -130,7 +130,7 @@ module StochasticDiffEq
          DRI1, DRI1NM, RI1, RI3, RI5, RI6, RDI1WM, RDI2WM, RDI3WM, RDI4WM,
          RS1, RS2,
          PL1WM, PL1WMA,
-         NON
+         NON, COM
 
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
It looks like there was a regression in the nonlinear solver. @kanav99 @YingboMa whoever has a free minute open first should take a look. This uses OrdinaryDiffEq's nonlinear solvers, so whatever changed there in the last 3-4 days would be the culprit.

frankschae:
I'll try to decrease the times for the new Stratonovich tests in the mean time

ChrisRackauckas:
https://github.com/SciML/StochasticDiffEq.jl/pull/345 should fix this.

frankschae:
That was a close one :D
 GROUP=WeakConvergence4 49 min 55 sec

PR #347, opened 2020-08-10
 [WIP] Levy area for non-commutative noise processes

frankschae:
I added tests for the iterated stochastic integrals (originally from https://github.com/SciML/StochasticDiffEq.jl/pull/209) and I also added the scheme by Kloeden, Platen, and Wright.

It's not yet passing the following test from https://github.com/SciML/StochasticDiffEq.jl/pull/268 :
```
sim6 = analyticless_test_convergence(dts,prob2,RKMil_General(),test_dt,trajectories=300, use_noise_grid=false)
@test abs(sim6.𝒪est[:final] - 0.5) < 0.2
@test_broken abs(sim6.𝒪est[:final] - 1.0) < 0.2
```

is there a good way of testing the iterated integrals directly, i.e., without reference to an algorithm like Milstein (for example maybe some moments)?


frankschae:
@mschauer  do you maybe know a good test for the approximation of the iterated integrals?

mschauer:
Does this help? https://math.stackexchange.com/questions/2906268/calculate-a-multiple-ito-integral

frankschae:
Following the notation in Kloeden, P. E., Platen, E., & Wright, I. W. (1992). The approximation of multiple stochastic integrals. Stochastic analysis and applications, 10(4), 431-441, for the definition of a multiple stochastic (Stratonovich) integral:
```
J_{(i,j), s, t} = \int_{s}^{t} \int_{s}^{s1} \int_{s}^{s2} dW^i_{s1} dW^j_{s2} 
```
They also state the result for the diagonal values -- as in the stackexchange discussion: 
```
J_{(i,i), \tau_n, \tau_{n+1} } =  1/2 (\delta W_n^i)^2
```
and say that the off-diagonal values cannot be expressed easily as functionals of `\delta W_n^{j1}` and `\delta W_n^{j2}` (which is also stated in Wiktorsson's paper on the simulation of iterated integrals for multiple independent Brownian motions)

I added the tests for the diagonal values already. However, I am wondering if there doesn't exist some test for the *off-diagonal* values to check if the approximation is correct (like statistics of the solution(?)) that doesn't rely on testing the convergence of the Milstein solver for non-commutative noise processes.

mschauer:
Maybe the product formula for iterated ito integrals gets you somewhere: the product of In(f) and Im(g) is a finite and explicit sum of k-multiple integrals, k < max(n, m), see https://www.sciencedirect.com/science/article/pii/S0304414997001014

ChrisRackauckas:
I rebased onto master so that tests shouldn't be effected by the v1.5 changes to code coverage.

You might want to make a dispatch that's the simplest iterated integral via just the series definition with a ton of terms, use that to get something working, and then make other definitions.

frankschae:
If I have a fixed `W.dW`
Shouldn't the estimation of the area converge as the truncation parameter is increased?
`StochasticDiffEq.get_iterated_I(dt, W.dW, KPWdiagoop, Int(1e6), 1)`

If I don't fix the seed between subsequent evaluations, e.g.,
```julia
Random.seed!(seed)
a = StochasticDiffEq.get_iterated_I(dt, W.dW, KPWdiagoop, Int(1e6), 1)
b = StochasticDiffEq.get_iterated_I(dt, W.dW, KPWdiagoop, Int(1e6), 1)
```

I obtain for `a-b`:
```julia
10×10 Array{Float64,2}:
  0.0         -0.0411685   -0.0289268   …   0.0502469   -0.055609
  0.0411685    0.0         -0.0640189        -0.0234345    0.0549249
  0.0289268    0.0640189    0.0               0.0371386    0.0380299
 -0.0327852   -0.0121136   -0.029224       -0.0293781   -0.0816394
 -0.0488158   -0.00869942  -0.0442117      -0.00291947  -0.032175
  0.108047    -0.122451    -0.147561    …     0.0472747    0.00938384
 -0.00435072  -0.0226749   -0.00495736       -0.0028539   -0.0372274
  0.0207632   -0.0704816   -0.084078         -4.17179e-5  -0.0992825
 -0.0502469    0.0234345   -0.0371386            0.0         -0.0859999
  0.055609    -0.0549249   -0.0380299            0.0859999    0.0
```

ChrisRackauckas:
Yes it should converge and yes that is odd.

frankschae:
Based on the following code I checked the convergence of the off-diagonal element J_{12}:
```julia
KPWdiagiip = StochasticDiffEq.KPWJ_iip(W.dW)

vals1 = []
for i in 1:10000
  Random.seed!(seed)
  StochasticDiffEq.get_iterated_I!(dt, W.dW, KPWdiagiip, Int(i), 1)
  push!(vals1,KPWdiagiip.WikJ[1,2])
end

using Plots; convplot = plot(vals1)
```
I find that the area and the iterated integral, respectively, converge if (and only if) I fix the random seed for each `i` which somehow makes sense to me since the relative contribution of the first random number seems to be very important because of the scaling by `k` in the loop:   `WikA += (kronprod1 - kronprod2)/k` (but of course it's weird since this implies that the area depends on the seed). This is shown in the first plot. 

The second plot shows the same as above but without fixing the seed, i.e., `#Random.seed!(seed)`

fixed_seed
without_fixed_seed

I can reproduce exactly the same behaviour when I use the implementation from this SDE python package:
https://github.com/mattja/sdeint -- though not very surprising as it looks like the function `def _Aterm(N, h, m, k, dW)` agrees with our function `get_iterated_I(dt, dW, Wik::KPWJ_oop, p=nothing, C=1)`.

frankschae:
Since we don't know the analytical solution for a non-commutative noise process, we can't use:
```julia
test_convergence(dts,prob,RKMil_General(),trajectories=1000)
```
but instead we are using
```julia
analyticless_test_convergence(dts,prob2,RKMil_General(),test_dt,trajectories=1000, use_noise_grid=false)
```
The analyticless tests are based on the simulation of the SDE with a smaller `dt` (`test_dt` above)  whereas the result of this is used as the true solution. To check for strong convergence, we need to re-use the noise from the true_solution for all other simulations with different `dt> test_dt`, see: 
https://github.com/SciML/DiffEqDevTools.jl/blob/b3a326f4ae0247d3b103b69d17eed604de5bfba6/src/convergence.jl#L102

I am wondering if we are just missing to re-use also the new random numbers drawn in the iterated integrals in the same way as we are re-using the values of `dW` .

ChrisRackauckas:
>I am wondering if we are just missing to re-use also the new random numbers drawn in the iterated integrals in the same way as we are re-using the values of dW .

Good point, that's probably the issue.

ChrisRackauckas:
Given how much is passing, I think it might be correct but I agree that the analyticless convergence testing might not be sufficiently storing enough of the randomness to prove it. I think it would be good to complete this PR by marking that as a failure for now, and showing that you at least get Order 1 using this whole method on https://github.com/SciML/StochasticDiffEq.jl/blob/master/test/commutative_tests.jl#L5-L33 . This all is enough to convince me that it's correct, though we should then open up an issue to someday complete the full convergence test on some non-commutative SDE.

src/solve.jl
```diff
@@ -323,23 +334,34 @@ function DiffEqBase.__init(
...
+            rand_prototype2 .= false
+          end
+          W = WienerProcess(t,rand_prototype,rand_prototype2,
                              save_everystep=save_noise,
                              rng = Xorshifts.Xoroshiro128Plus(_seed))
+        elseif alg===RKMil_General()
+          m = length(rand_prototype)
+          if typeof(rand_prototype) <: Number || alg.p == nothing
+            rand_prototype2 = nothing
           else
```
ChrisRackauckas, line 353: This whole thing is very messy and not general: it should definitely get a generalized form.

ChrisRackauckas:
🎉 this is great!

Issue #351, opened 2020-08-26
Type unstable evaluation when computing the derivative of a solution object 


rvignolo:
This is a MWE:
```julia
julia> using DifferentialEquations

julia> function f1!(du, u, p, t) 
         du[1] = t * u[1]
       end
f1! (generic function with 1 method)

julia> function g1!(du, u, p, t)
         du[1] = 0
       end
g1! (generic function with 1 method)

julia> prob = SDEProblem(f1!, g1!, ones(1), (0., 1.))
SDEProblem with uType Array{Float64,1} and tType Float64. In-place: true
timespan: (0.0, 1.0)
u0: [1.0]

julia> sol = solve(prob);

julia> const solution = sol;

julia> using ForwardDiff

julia> ForwardDiff.derivative(t -> solution(t, idxs = 1), 0.5)
0.5625617554051718

julia> @code_warntype ForwardDiff.derivative(t -> solution(t, idxs = 1), 0.5)
Variables
  #self#::Core.Compiler.Const(ForwardDiff.derivative, false)
  f::Core.Compiler.Const(var"#17#18"(), false)
  x::Float64
  T::Type{ForwardDiff.Tag{var"#17#18",Float64}}
Body::Float64
1 ─      nothing
│   %2 = ForwardDiff.Tag(f, $(Expr(:static_parameter, 2)))::Core.Compiler.Const(ForwardDiff.Tag{var"#17#18",Float64}(), false)
│        (T = ForwardDiff.typeof(%2))
│   %4 = T::Core.Compiler.Const(ForwardDiff.Tag{var"#17#18",Float64}, false)::Core.Compiler.Const(ForwardDiff.Tag{var"#17#18",Float64}, false)
│   %5 = Core.apply_type(ForwardDiff.Dual, T::Core.Compiler.Const(ForwardDiff.Tag{var"#17#18",Float64}, false))::Core.Compiler.Const(ForwardDiff.Dual{ForwardDiff.Tag{var"#17#18",Float64},V,N} where N where V, false)
│   %6 = ForwardDiff.one(x)::Core.Compiler.Const(1.0, false)
│   %7 = (%5)(x, %6)::Core.Compiler.PartialStruct(ForwardDiff.Dual{ForwardDiff.Tag{var"#17#18",Float64},Float64,1}, Any[Float64, Core.Compiler.Const(Partials(1.0,), false)])
[...]

PR #352, opened 2020-08-26
Fix type inference of interpolation

devmotion:
Fixes https://github.com/SciML/StochasticDiffEq.jl/issues/351.

That's the good thing. The bad thing is that the implementation is just copied from OrdinaryDiffEq.

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -103,44 +103,44 @@ sde_interpolation(tvals,ts,timeseries,ks)
 Get the value at tvals where the solution is known at the
 times ts (sorted), with values timeseries and derivatives ks
 """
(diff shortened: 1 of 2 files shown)


rvignolo:
Hi, thanks for this! 

I was wondering, what are the differences between the `OrdinaryDiffEq._searchsortedlast` and `Base.searchsortedlast`?

ChrisRackauckas:
`OrdinaryDiffEq._searchsortedlast` was a lot faster IIRC

rvignolo:
thank you both.

ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/blob/master/src/interpolation.jl is the last spot where this kind of code also exists. We should try to get it all merged to the same system, but it's never been the highest priority.

PR #360, opened 2020-09-17
remove generate_weak_solution and instead use test_convergence

frankschae:
Rewrote the weak tests such that they now use the `test_convergence` function from

https://github.com/SciML/DiffEqDevTools.jl/pull/74


ChrisRackauckas:
Great to see all green checkmarks 😄

PR #370, opened 2020-12-08
Use heap initializations defined in OrdinaryDiffEq

devmotion:
This PR removes the initialization methods for `tstops`, `saveat`, and `d_discontinuities` which were duplicates of the implementation in OrdinaryDiffEq and instead uses the new methods in OrdinaryDiffEq, introduced in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1307.


ChrisRackauckas:
oh great, always a sign of a good refactor.

Issue #372, opened 2020-12-21
RKMilCommute is missing a dispatch for OOP


ChrisRackauckas:

ChrisRackauckas:
Migrated to OrdinaryDiffEq.jl as part of StochasticDiffEq.jl archival. See the corresponding issue at https://github.com/SciML/OrdinaryDiffEq.jl/issues?q=is%3Aissue+label%3ASDE

PR #383, opened 2021-01-11
RKMilCommute constant cache perform step

frankschae:
Solves #372 .

Also fixes two small bugs for `RKMilGeneral`.
Additionally, The iip_weak test file used a pre-defined SDE problem from DiffEqProblemLibrary before (which, however, is defined as oop). This is now also corrected.

### DIFF
--- a/src/perform_step/low_order.jl
+++ b/src/perform_step/low_order.jl
@@ -249,21 +249,18 @@ end
 @muladd function perform_step!(integrator,cache::RKMilCommuteConstantCache,f=integrator.f)
   @unpack t,dt,uprev,u,W,p = integrator
   dW = W.dW; sqdt = integrator.sqdt
+  Wik = cache.WikJ
 
   ggprime_norm = 0.0
 
+  WikJ = get_iterated_I(dt, dW, W.dZ, Wik)
+
   mil_correction = zero(u)
   if alg_interpretation(integrator.alg) == :Ito
-    if typeof(dW) <: Number
-      WikJ = 0.5 .* ((dW) .* (dW)'  .- dt)
+    if typeof(dW) <: Number || is_diagonal_noise(integrator.sol.prob)
+      WikJ = WikJ .- 1//2 .* dt
     else
-      WikJ = 0.5 .* (vec(dW) .* vec(dW)'  .- dt .* Eye{eltype(W.dW)}(length(W.dW)))
-    end
-  else
-    if typeof(dW) <: Number
-      WikJ = 0.5 * dW .* dW'
-    else
-      WikJ = 0.5 .* vec(dW) .* vec(dW)'
+      WikJ -= 1//2 .* UniformScaling(dt)
     end
   end
 
@@ -271,27 +268,35 @@ end
   L = integrator.g(uprev,p,t)
 
   K = uprev + dt*du1
-  for j = 1:length(dW)
-    if typeof(dW) <: Number
-      Kj = K + sqdt*L
-    else
-      Kj = K + sqdt*@view(L[:,j])
-    end
-    gtmp = integrator.g(Kj,p,t)
-    Dgj = (gtmp - L)/sqdt
-    if integrator.opts.adaptive
-        ggprime_norm += integrator.opts.internalnorm(Dgj,t)
-    end
-    if typeof(dW) <: Number
-      tmp = Dgj*WikJ
-    else
-      tmp = Dgj*@view(WikJ[:,j])
-    end
-    mil_correction += tmp
-  end
-  tmp = L*dW
-  u = uprev + dt*du1 + tmp + mil_correction
 
+  if is_diagonal_noise(integrator.sol.prob)
+    tmp = (alg_interpretation(integrator.alg) == :Ito ? K : uprev) .+ integrator.sqdt .* L
+    gtmp = integrator.g(tmp,p,t)
+    Dgj = (gtmp - L)/sqdt
+    ggprime_norm = integrator.opts.internalnorm(Dgj,t)
+    u = @.. K + L*dW + Dgj*WikJ
+  else
+    for j = 1:length(dW)
+      if typeof(dW) <: Number
+        Kj = K + sqdt*L
+      else
+        Kj = K + sqdt*@view(L[:,j])
+      end
+      gtmp = integrator.g(Kj,p,t)
+      Dgj = (gtmp - L)/sqdt
+      if integrator.opts.adaptive
+        ggprime_norm += integrator.opts.internalnorm(Dgj,t)
+      end
+      if typeof(dW) <: Number
+        tmp = Dgj*WikJ
+      else
+        tmp = Dgj*@view(WikJ[:,j])
+      end
+      mil_correction += tmp
+    end
+    tmp = L*dW
+    u = uprev + dt*du1 + tmp + mil_correction
+  end
   if integrator.opts.adaptive
       En = integrator.opts.internalnorm(dW,t)^3*ggprime_norm^2 / 6
       du2 = integrator.f(K,p,t+dt)
@@ -313,30 +318,46 @@ end
 
   ggprime_norm = 0.0
 
+  Wik = cache.WikJ
+
+  get_iterated_I!(dt, dW, W.dZ, Wik)
+  WikJ = Wik.WikJ
+
   @.. mil_correction = zero(u)
   if alg_interpretation(integrator.alg) == :Ito
-    WikJ .= 0.5 .* (vec(dW) .* vec(dW)'  .- dt .* Eye{eltype(W.dW)}(length(W.dW)))
-  else
-    WikJ .= 0.5 .* vec(dW) .* vec(dW)'
+    if typeof(dW) <: Number || is_diagonal_noise(integrator.sol.prob)
+      @.. WikJ -= 1//2*dt
+    else
+      WikJ -= 1//2 .* UniformScaling(dt)
+    end
   end
 
   integrator.f(du1,uprev,p,t)
   integrator.g(L,uprev,p,t)
 
   @.. K = uprev + dt*du1
-  for j = 1:length(dW)
-    @.. Kj = K + sqdt*@view(L[:,j]) # This works too
-    #Kj .= uprev .+ sqdt*L[:,j]
-    integrator.g(gtmp,Kj,p,t)
+
+  if is_diagonal_noise(integrator.sol.prob)
+    tmp .= (alg_interpretation(integrator.alg) == :Ito ? K : uprev) .+ integrator.sqdt .* L
+    integrator.g(gtmp,tmp,p,t)
     @.. Dgj = (gtmp - L)/sqdt
-    if integrator.opts.adaptive
-        ggprime_norm += integrator.opts.internalnorm(Dgj,t)
+    ggprime_norm = integrator.opts.internalnorm(Dgj,t)
+    @.. u = K + L*dW + Dgj*WikJ
+  else
+    for j = 1:length(dW)
+      @.. Kj = K + sqdt*@view(L[:,j]) # This works too
+      #Kj .= uprev .+ sqdt*L[:,j]
+      integrator.g(gtmp,Kj,p,t)
+      @.. Dgj = (gtmp - L)/sqdt
+      if integrator.opts.adaptive
+          ggprime_norm += integrator.opts.internalnorm(Dgj,t)
+      end
+      mul!(tmp,Dgj,@view(WikJ[:,j]))
+      mil_correction .+= tmp
     end
-    mul!(tmp,Dgj,@view(WikJ[:,j]))
-    mil_correction .+= tmp
+    mul!(tmp,L,dW)
+    @.. u .= uprev + dt*du1 + tmp + mil_correction
   end
-  mul!(tmp,L,dW)
-  @.. u .= uprev + dt*du1 + tmp + mil_correction
 
   if integrator.opts.adaptive
       En = integrator.opts.internalnorm(W.dW,t)^3*ggprime_norm^2 / 6
--- a/src/perform_step/low_order.jl
+++ b/src/perform_step/low_order.jl
@@ -246,12 +246,70 @@ end
   end
 end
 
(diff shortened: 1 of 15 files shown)


src/perform_step/low_order.jl
```diff
@@ -269,7 +327,7 @@ end
   for j = 1:length(dW)
     @.. Kj = K + sqdt*@view(L[:,j]) # This works too
```
ChrisRackauckas, line 328: it wants this to be `@.`? Seems a bit odd.

frankschae: I think it will need an `if is_diagonal_noise(integrator.sol.prob)` branch, similar to the one from `RKMilGeneral()`. I think the error essentially is due to the matrix-like form of u0.

ChrisRackauckas: Oh yes. `@..` specializes in a way that requires a 1-1 map.

test/sde/sde_twodimlinear_tests.jl
```diff
@@ -103,10 +107,13 @@ sim2 = test_convergence(dts,prob,WangLi3SMil_F(),trajectories=100)
...
 print(".")
 
+Random.seed!(100)
 eigen_est = (integrator) -> integrator.eigen_est = 10.0
 for Alg in [SROCK1, SROCK2], alg in [Alg(), Alg(eigen_est=eigen_est)]
-  sim2 = test_convergence(dts,prob,alg,trajectories=100)
+  local sim2
+  sim2 = test_convergence(dts,prob,alg,trajectories=150)
   @test abs(sim2.𝒪est[:l∞]-1) < 0.2
+  @show sim2.𝒪est[:l∞]
```
frankschae, line 116: It's failing here (0.22 instead of <0.2  with the old version using `trajectories =100` and no re-seeding). Locally, the tests are actually passing on my machine.

ChrisRackauckas: ehh seems like just random variation

ChrisRackauckas: 150 trajectories is still pretty low

test/weak_convergence/iip_weak.jl
```diff
@@ -40,9 +52,9 @@ sim = test_convergence(dts,prob,SROCK1(),save_everystep=false,trajectories=Int(4
 #@test abs(sim.𝒪est[:weak_l2]-1) < 0.3
 #@test abs(sim.𝒪est[:weak_l∞]-1) < 0.3
 println("SROCK2")
-sim = test_convergence(dts,prob,SROCK2(),save_everystep=false,trajectories=Int(4e4),
+sim = test_convergence(dts,prob,SROCK2(),save_everystep=false,trajectories=Int(1e5),
                         weak_timeseries_errors=false)
-@test abs(sim.𝒪est[:weak_final]-1.5) < 0.3
+@test abs(sim.𝒪est[:weak_final]-1.5) < 0.31
```
ChrisRackauckas, line 57: seems like a real test error here? Odd...

ChrisRackauckas: Feel free to move tests into the BuildKite queue to speed them up. We can make them be highly multithreaded and such if we need to.

frankschae: Do the weak adaptive tests already run on the fastest GPUs/CPUs (buildkite/stochasticdiffeq-dot-jl/weakconvergen..)? Looks like they got aborted because of their runtime.

ChrisRackauckas: Yes they were on one of the fast machines with a GPU. Were they extended?

frankschae: Not yet..

Should I just try increasing
https://github.com/SciML/StochasticDiffEq.jl/blob/a96819fca0fccf29f4210e679bb36b6e7c50ef64/.buildkite/pipeline.yml#L32
if they fail again or is it better to split the file/try to make the tests faster?

ChrisRackauckas: yeah double it and see what happens.

ChrisRackauckas:
Looks like the amount of time is just over 2 hours.

ChrisRackauckas:
WeakConvergence just needs CPU right? So we can put it in a different queue. All of the tests are already multithreaded?

frankschae:
> All of the tests are already multithreaded?

Yes, I think all tests are already multithreaded, since they should all use `EnsembleThreads()` from
https://github.com/SciML/DiffEqDevTools.jl/blob/master/src/convergence.jl

> WeakConvergence just needs CPU right? So we can put it in a different queue. 

Yes.

ChrisRackauckas:
https://buildkite.com/julialang/stochasticdiffeq-dot-jl/builds/40#52ddc367-5e7c-4deb-a320-193cf981b835/359-869

Looks like a random failure. FWIW, it's great that we're finding out how to make these stable and not rely on just a seed. On another note, damn these take a lot of compute. I think we will want a semi-SIMD form @chriselrod mentioned before.

I'll look at what's going on with that GPU test. I think that's a regression.

frankschae:
yeah, indeed the tests get really expensive .. 
 
> I'll look at what's going on with that GPU test. I think that's a regression.
Yes.. I also wanted to ask you already about that. I thought it must be something almost trivial as the tests were passing before we added the weak convergence tests to the buildkite queue

ChrisRackauckas:
@vchuravy @jpsamaroo could you take a look at https://buildkite.com/julialang/stochasticdiffeq-dot-jl/builds/40#de077c9c-e46e-4a31-b7ac-1fdd10b45bf4/359-733 and see if you know what that is? I am confused since I can't recreate it. I locally grabbed the test and it ran fine:

```julia
using StochasticDiffEq, Test, Random
using DiffEqGPU
using CUDA

function weak_error(prob,alg,numtraj, batchsize, f_true,trange;abstol=1,reltol=0,ensemblealg=EnsembleCPUArray())
  sol = @time solve(prob,alg,ensemblealg,
    dt=0.01f0,adaptive=true,abstol=abstol,reltol=reltol,
    trajectories=numtraj,batch_size=batchsize,
    saveat = trange
    )
  computed_exp = (sol.u/numtraj)[1,:]
  true_exp = f_true.(trange)
  return sum((computed_exp-true_exp).^2)/length(trange)
end

function prob_func(prob, i, repeat)
    #(i%50000 == 0) && @show i
    remake(prob,seed=seeds[i])
end

function reduction(u,batch,I)
  u.+sum(batch),false
end

function output_func(sol,i)
  #h1(asinh(sol[end][1])),false
  h1.(asinh.(sol)),false
end

# prob 1
u₀ = [0.0f0]
tspan = (0.0f0,2.0f0)
h1(z) = z^3-6*z^2+8*z

function f1!(du,u,p,t)
 @inbounds begin
     du[1] = 1//2*u[1]+sqrt(u[1]^2 +1)
 end
 nothing
end

function g1!(du,u,p,t)
 @inbounds begin
     du[1] = sqrt(u[1]^2 +1)
 end
 nothing
end

f_true1(t) = t^3-3*t^2+2*t

prob1 = SDEProblem(f1!,g1!,u₀,tspan)
ensemble_prob1 = EnsembleProblem(prob1;
        output_func = output_func,
        prob_func = prob_func,
        reduction = reduction,
        u_init=Vector{eltype(prob1.u0)}([0.0])
        )


# prob 2
u₀ = [0.1f0,0.1f0]
function f2!(du,u,p,t)
  @inbounds begin
    du[1] = 3//2*u[1]
    du[2] = 3//2*u[2]
  end
  nothing
end
function g2!(du,u,p,t)
  @inbounds begin
    du[1] = 1//10*u[1]
    du[2] = 1//10*u[2]
  end
  nothing
end

f_true2(t) = 1//10*exp(3//2*t) #1//100*exp(301//100*t)

h2(z) = z #z^2

prob2 = SDEProblem(f2!,g2!,u₀,tspan)
ensemble_prob2 = EnsembleProblem(prob2;
        output_func = (sol,i) -> (h2.(sol),false),
        prob_func = prob_func,
        reduction = reduction,
        u_init=Vector{eltype(prob2.u0)}([0.0, 0.0])
        )



tsave = 0.0f0:0.05f0:2f0



probs = Vector{EnsembleProblem}(undef, 2)
probs[1] = ensemble_prob1
probs[2] = ensemble_prob2

ftrue = Vector{}(undef, 2)
ftrue[1] = f_true1
ftrue[2] = f_true2

numtraj = Int(1e5)
seed = 100
Random.seed!(seed)
seeds = rand(UInt, numtraj)

numtraj = Int(1e5)
seed = 100
Random.seed!(seed)
seeds = rand(UInt, numtraj)

for i in 1:2
  @show i

  err1 = weak_error(probs[i],DRI1NM(),numtraj,Int(1e4),ftrue[i],tsave,abstol=1f0,reltol=1f0, ensemblealg=EnsembleGPUArray())
  @show err1
  # err2 = weak_error(probs[i],DRI1NM(),numtraj,Int(1e4),ftrue[i],tsave,abstol=0.1f0,reltol=0.1f0, ensemblealg=EnsembleGPUArray())
  # @show err2
  err3 = weak_error(probs[i],DRI1NM(),numtraj,Int(1e4),ftrue[i],tsave,abstol=0.01f0,reltol=0.01f0, ensemblealg=EnsembleGPUArray())
  @show err3
  @test err1 > err3
  println("")
end
```

chriselrod:
> FWIW, it's great that we're finding out how to make these stable and not rely on just a seed. On another note, damn these take a lot of compute. I think we will want a semi-SIMD form @chriselrod mentioned before.

What do you want to SIMD?


Also, I could revive the PR on using `VectorizedRNG`. I've added a scalar SIMD RNG:
```julia
julia> x = Vector{Float64}(undef, 1024);

julia> using VectorizedRNG, Random

julia> @benchmark randn!(local_rng(), $x)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.298 μs (0.00% GC)
  median time:      1.344 μs (0.00% GC)
  mean time:        1.347 μs (0.00% GC)
  maximum time:     2.637 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> @benchmark randn!($x)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.062 μs (0.00% GC)
  median time:      2.176 μs (0.00% GC)
  mean time:        2.182 μs (0.00% GC)
  maximum time:     3.321 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     9

julia> @benchmark randn(local_rng())
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.695 ns (0.00% GC)
  median time:      3.050 ns (0.00% GC)
  mean time:        3.029 ns (0.00% GC)
  maximum time:     5.762 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark randn()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.353 ns (0.00% GC)
  median time:      6.712 ns (0.00% GC)
  mean time:        6.736 ns (0.00% GC)
  maximum time:     23.688 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
But this computer has AVX512.
After the recent performance improvements in the base `randn!`, it might not actually be faster anymore without it.
If `randn(local_rng())` isn't faster without avx512, switching to the Ziggurat algorithm base `randn()` uses probably would make it faster because `rand` holds a much larger advantage:
```julia
julia> @benchmark rand!(local_rng(), $x)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     168.105 ns (0.00% GC)
  median time:      168.276 ns (0.00% GC)
  mean time:        168.450 ns (0.00% GC)
  maximum time:     192.335 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     750

julia> @benchmark rand!($x)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     594.022 ns (0.00% GC)
  median time:      597.654 ns (0.00% GC)
  mean time:        598.793 ns (0.00% GC)
  maximum time:     701.682 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     179

julia> @benchmark rand(local_rng())
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.855 ns (0.00% GC)
  median time:      1.927 ns (0.00% GC)
  mean time:        1.931 ns (0.00% GC)
  maximum time:     5.347 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark rand()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     4.581 ns (0.00% GC)
  median time:      5.086 ns (0.00% GC)
  mean time:        5.099 ns (0.00% GC)
  maximum time:     17.237 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
and is thus probably faster on everything with AVX2.

But I wouldn't be surprised if RandomNumbers.jl wins if all you need is one scalar at a time.

ChrisRackauckas:
It's two pieces. One is the RNG. The other is that this is solving a ton of trajectories multithreaded, so instead of doing every one separate, we should do that semi-clumping we discussed before.

ChrisRackauckas:
We almost never need one scalar at time, so if you wanted to revive that effort for VectorizedRNGs I would be super happy.

ChrisRackauckas:
What's blocking this?

frankschae:
Mainly, the CUDA issue in buildkite/stochasticdiffeq-dot-jl/weakadaptive .

vchuravy:
I will take a look at the failure. It seems a change to adapt has broken `@Const`.

ChrisRackauckas:
I'm going to merge this to separate the concerns. But let's definitely continue this thread. I'll open two issues on the two points.

PR #391, opened 2021-01-25
Fix `handle_callback_modifiers!`

devmotion:
The current implementation seems wrong (see https://github.com/SciML/StochasticDelayDiffEq.jl/pull/35#discussion_r564026139).

It seems it is not tested currently and hence a test should be added. What would be a simple example that fails currently?

### DIFF
--- a/src/integrators/integrator_utils.jl
+++ b/src/integrators/integrator_utils.jl
@@ -260,16 +260,19 @@ end
   end
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
It would be a callback in https://github.com/SciML/StochasticDiffEq.jl/blob/master/test/tau_leaping.jl since the `P` term only applies to tau-leaping.

Issue #259, opened 2020-01-13
High weak order methods for Hamiltonian systems


ChrisRackauckas:
https://arxiv.org/abs/1304.3269

jamesgardner1421:
I'd be interested in implementing the BAOAB algorithm from the reference here. Could you provide some instruction on how best to begin? I've looked at the general developer documentation but wanted to check with you first in case there are any extra specifics I should consider.

ChrisRackauckas:
Yeah the devdocs are a good place to start to get a good feel for how it's all constructed:

http://devdocs.sciml.ai/latest/contributing/adding_algorithms/

Though this kind of system will require a partitioned system, so it might need something similar to DynamicalODEProblem for SDEs to be implemented in SciMLBase.jl. So the symplectic integrators for ODE systems + the devdocs is a good place to look, and from there, take a stab at it and I'll help you along the way.

HadrienNU:
I will be interested to implement the remaining algorithm in the paper (ABOBA and OBABO at least).

I was thinking to implement as well the GJ set of integrator [https://arxiv.org/abs/1909.04380v1](https://arxiv.org/abs/1909.04380v1) , if this is of interest

ChrisRackauckas:
That would be great to have, and definitely of interest.

ChrisRackauckas:
Migrated to OrdinaryDiffEq.jl as part of StochasticDiffEq.jl archival. See https://github.com/SciML/OrdinaryDiffEq.jl/issues/3183

PR #397, opened 2021-02-05
Add BAOAB algorithm

jamesgardner1421:
I've made a start on adding the BAOAB algorithm mentioned in issue #259. It relies upon the corresponding pull request I've submitted to SciMLBase that contains the new DynamicalSDEProblem/Function.

Currently I've been able to implement both the inplace and out-of-place versions and both give the same result.

I'm not sure that I've set up the caches/initialisation correctly with the prototypes and such but it works for now.

I've assumed a specific structure of the SDE with:
`dq = v*dt`
`dv = f1*dt - gamma*v*dt + g*sqrt(2gamma)*dW`
When f1 is the force, and g is sqrt(kT/m) this is equivalent to Langevin dynamics.
It's possible to make it more general and instead have
`dv = f1*dt - gamma*v*dt + g*dW`.
If you think this form is better I can change it.

Let me know how I should modify things to fit in with everything else.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -95,7 +95,7 @@ alg_order(alg::SMEB) = 1//1
 alg_order(alg::TauLeaping) = 1//1
 alg_order(alg::CaoTauLeaping) = 1//1
 
-alg_order(alg::BAOAB) = 2//1
+alg_order(alg::BAOAB) = 1//1
 
 alg_order(alg::SKenCarp) = 2//1
 alg_order(alg::Union{StochasticDiffEqCompositeAlgorithm,StochasticDiffEqRODECompositeAlgorithm}) = maximum(alg_order.(alg.algs))
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -95,6 +95,8 @@ alg_order(alg::SMEB) = 1//1
 alg_order(alg::TauLeaping) = 1//1
 alg_order(alg::CaoTauLeaping) = 1//1
 
+alg_order(alg::BAOAB) = 1//1 # Not sure what order it is
+
 alg_order(alg::SKenCarp) = 2//1
 alg_order(alg::Union{StochasticDiffEqCompositeAlgorithm,StochasticDiffEqRODECompositeAlgorithm}) = maximum(alg_order.(alg.algs))
 get_current_alg_order(alg::StochasticDiffEqAlgorithm,cache) = alg_order(alg)
@@ -196,6 +198,7 @@ alg_compatible(prob::DiffEqBase.AbstractSDEProblem,alg::RKMilGeneral) = true
 alg_compatible(prob::DiffEqBase.AbstractSDEProblem,alg::IIF1M) = true
 alg_compatible(prob::DiffEqBase.AbstractSDEProblem,alg::IIF2M) = true
 alg_compatible(prob::DiffEqBase.AbstractSDEProblem,alg::Union{StochasticDiffEqCompositeAlgorithm,StochasticDiffEqRODECompositeAlgorithm}) = max((alg_compatible(prob,a) for a in alg.algs)...)
+alg_compatible(prob::DiffEqBase.AbstractSDEProblem,alg::BAOAB) = is_diagonal_noise(prob)
 
 function alg_compatible(prob::JumpProblem,alg::Union{StochasticDiffEqJumpAdaptiveAlgorithm,StochasticDiffEqJumpAlgorithm})
     prob.prob isa DiscreteProblem
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -95,7 +95,7 @@ alg_order(alg::SMEB) = 1//1
 alg_order(alg::TauLeaping) = 1//1
 alg_order(alg::CaoTauLeaping) = 1//1
 
-alg_order(alg::BAOAB) = 1//1 # Not sure what order it is
+alg_order(alg::BAOAB) = 2//1
 
 alg_order(alg::SKenCarp) = 2//1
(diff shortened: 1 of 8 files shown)


src/alg_utils.jl
```diff
@@ -95,6 +95,8 @@ alg_order(alg::SMEB) = 1//1
 alg_order(alg::TauLeaping) = 1//1
 alg_order(alg::CaoTauLeaping) = 1//1
 
+alg_order(alg::BAOAB) = 1//1 # Not sure what order it is
```
ChrisRackauckas, line 98: that would be good to look into

test/sde/sde_dynamical.jl
```diff
@@ -0,0 +1,25 @@
+
```
ChrisRackauckas, line 1:
```suggestion
```

test/sde/sde_dynamical.jl
```diff
@@ -0,0 +1,25 @@
...
+using StochasticDiffEq, DiffEqNoiseProcess, Test, DiffEqDevTools
```
ChrisRackauckas, line 2: needs to be added to runtests.jl

ChrisRackauckas:
Needs a lower bound on SciMLBase 1.7

test/sde/sde_dynamical.jl
```diff
@@ -0,0 +1,22 @@
...
+sol1 = solve(prob,BAOAB(gamma=γ);dt=1/10,save_noise=true)
+
+f1_harmonic_iip(dv,v,u,p,t) = dv .= f1_harmonic(v,u,p,t)
+f2_harmonic_iip(du,v,u,p,t) = du .= f2_harmonic(v,u,p,t)
+g_iip(du,u,p,t) = du .= g(u,p,t)
+
+prob = DynamicalSDEProblem(f1_harmonic_iip,f2_harmonic_iip,g_iip,v0,u0,(0.0,5.0); noise=NoiseWrapper(sol1.W))
+
+sol2 = solve(prob,BAOAB(gamma=γ);dt=1/10)
+@test sol1[:] ≈ sol2[:]
```
ChrisRackauckas, line 22: this should get convergence tests

test/sde/sde_dynamical.jl
```diff
@@ -0,0 +1,31 @@
...
+
+prob2 = DynamicalSDEProblem(f1_harmonic_iip,f2_harmonic_iip,g_iip,v0,u0,(0.0,5.0); noise=NoiseWrapper(sol1.W))
+sol2 = solve(prob2,BAOAB(gamma=γ);dt=1/10)
+
+@test sol1[:] ≈ sol2[:]
+
+dts = (1/2) .^ (9:-1:3)
+# Can't use NoiseGrid as noise is not generated with the correct size in convergence.jl. We require noise with shape of v.
+# I don't think there are any analytic solutions for problems of this type
+sim1  = analyticless_test_convergence(dts,prob1,BAOAB(gamma=γ),(1/2)^10;trajectories=Int(1e2),use_noise_grid=false)
```
ChrisRackauckas, line 28: That's definitely not enough trajectories to produce a weak convergence estimate. Do a local plot with like 5e5 trajectories and show me the `plot(sim1)`.

jamesgardner1421: errors

Here's the plot with 5e4 trajectories with tspan=(0.0,0.5). I seemed to be running into memory issues if I try more trajectories or a longer tspan. However, it does seem to be converged now. I also tried with 1e4 trajectories and the plot is only marginally different.

Here's the estimate I obtain:
```julia
julia> sim1.𝒪est
Dict{Any,Any} with 2 entries:
  :weak_final => 1.45594
  :final      => 0.802602
```

ChrisRackauckas: What if you make the dts a bit smaller? the strong convergence estimate seems to not entire its asymtopic regime until half way through.

jamesgardner1421: errors

This is with only 1e4 trajectories as it takes a very long time with smaller timesteps. The test dt here is 0.5^12.
```julia
julia> sim1.𝒪est
Dict{Any,Any} with 2 entries:
  :weak_final => 1.19034
  :final      => 0.777283
```

This doesn't quite match up with the previous plot however so I guess it's not properly converged. I'm not sure I can run it for many more trajectories though.

ChrisRackauckas: More trajectories won't help. You'd see that because it would flatline in a noisy fashion on the left end due to not having enough values for the sampling error. This doesn't show sampling eror. It shows 1st order convergence.

ChrisRackauckas: Lower `dt`s would be the check to get rid of that wobble though.

jamesgardner1421: Oh ok, does this mean the algorithm is 1st order then? I don't know much about SDEs, I was just going off the plots in the paper, but that's converging observables calculated from the distributions obtained from the integration. Is that different?

ChrisRackauckas: A lot of the convergence diagrams aren't well-behaved in their paper either. Can you try and recreate one and see if it's around the same?

jamesgardner1421: It turns out I had made a mistake in the algorithm which I've now fixed. The convergence is a lot more well behaved now and I was able to roughly reproduce one of their plots.
image
image

I did 2000 trajectories with 1e6 steps and they did 2000 with 1e8 so I think it's safe to assume I'd converge to the same result given enough timesteps.

ChrisRackauckas (approved):
awesome!

PR #409, opened 2021-03-29
pass `t` to reinit!

frankschae:
If we have. e.g., a long NoiseGrid (with sol.W.t[end] > sol.t[end]) and reverse it, we don't want to start at the initial time of the NoiseGrid but at the time corresponding to sol.t[end].

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -405,7 +405,7 @@ function DiffEqBase.__init(
     W = prob.noise
     if hasfield(typeof(W),:t) && W.reset
       if W.t[end] != t
-        reinit!(W,t)
+        reinit!(W,t, t0=t)
       end
       # Reseed
       if typeof(W) <: NoiseProcess && W.reseed


ChrisRackauckas (approved):

ChrisRackauckas:
Looks like there's a test failure?

ChrisRackauckas:
Is the weakadaptive test actually using GPUs or should that move to pure CPU CI?

frankschae:
it has a GPU test but IIRC that was mainly to test the implementation of the `DRI1NM()` solver because in an earlier version of the solver, scalar indexing was used (which made it very slow on the GPU).

ChrisRackauckas:
We should split the GPU script out so that not as much time is required on the GPU machines.

ChrisRackauckas:
https://buildkite.com/julialang/stochasticdiffeq-dot-jl/builds/103#2513167d-9383-45bd-88be-702b1a05a843/386-733

This error is fixed by KernelAbstraction v0.5.5 which will be merged shortly.

ChrisRackauckas:
Seems to be the same weak adaptive test failing every time? 

https://buildkite.com/julialang/stochasticdiffeq-dot-jl/builds/111#b1da625a-25a0-4257-8c64-708102322115/386-575

Seems like that might not just be tolerance/trajectories issue./

PR #417, opened 2021-05-19
controllers

ranocha:
Companion PR to SciML/OrdinaryDiffEq.jl#1384. Locally, all tests pass for me.


ChrisRackauckas:
Does StochasticDelayDiffEq need updates?

ranocha:
> StochasticDelayDiffEq

Maybe, I didn't know about that :sweat_smile:

ranocha:
Yeah, it will need some updates, at least for the options

ChrisRackauckas:
The remaining convergence tests don't use the controllers.

ChrisRackauckas:
@frankschae some of those DiffEqNoiseProcess tolerances are a little low:

https://github.com/SciML/StochasticDiffEq.jl/runs/2621672149#step:7:321

1e-16 won't necessarily be resolvable as a relative tolerance with Float64.

frankschae:
I see.. I made a note to bump the tolerances for them.

PR #419, opened 2021-05-23
Improve controller typing

ChrisRackauckas:
Matches https://github.com/SciML/OrdinaryDiffEq.jl/pull/1419

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -484,7 +484,7 @@ function DiffEqBase.__init(
   end
 
   if controller === nothing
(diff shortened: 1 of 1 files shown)


src/solve.jl
```diff
@@ -484,7 +484,9 @@ function DiffEqBase.__init(
   end
 
   if controller === nothing
-    controller = default_controller(alg, cache, QT(qoldinit), beta1, beta2)
+    controller = default_controller(_alg, cache, convert(QT,qoldinit),
```
YingboMa, line 487:
```suggestion
    controller = default_controller(alg, cache, convert(QT,qoldinit),
```

Issue #427, opened 2021-07-22
Weak adaptive solvers give rise to Out-of-memory errors


WouterJRV:
See https://discourse.julialang.org/t/why-does-my-devectorized-code-keep-allocating-memory/62727/11 

For a set of Order 1e4 coupled SDE equations with additive diagonal noise, SOSRA  and similar works fine, but the DRI-methods crash with OutOfMemoryError() 

Apparently, this is "because Ihat2 uses length(u)^2 memory for the pairwise 3-point random numbers"

ChrisRackauckas:
@frankschae is there a good way to avoid allocating the matrix of all pairwise random numbers there?

frankschae:
yeah we never use the full matrix `Ihat2` but only its entries in `perform_step!`. These (scalar) noise increments are only used in the last stage value in all SRK schemes IIRC. So, one option is probably to compute them on the fly when they are needed. I tried to understand if we can avoid computing them for diagonal processes. Perhaps, only the diagonal of the matrix is sufficient in that case anyway.

WouterJRV:
With StochasticDiffEq v6.36.0 it should be working, right? Unfortunately, I'm still getting the same error.

ChrisRackauckas:
It can't be the same error because the line it was pointing to doesn't exist anymore. Maybe similar error different stack trace?

WouterJRV:
`ERROR: LoadError: OutOfMemoryError()
Stacktrace:
 [1] Array at ./boot.jl:408 [inlined]
 [2] Array at ./boot.jl:416 [inlined]
 [3] zeros at ./array.jl:525 [inlined]
 [4] zeros at ./array.jl:521 [inlined]
[...]

PR #428, opened 2021-07-27
Remove Ihat2 matrix from weak solvers

frankschae:
see https://github.com/SciML/StochasticDiffEq.jl/issues/427 

`NON2` is the only method that doesn't exclusively access individual elements of the matrix `Ihat2` but instead it also contains matmuls of `g` with rows of `Ihat2`.


ChrisRackauckas (approved):
This is better for GPUs too. Nice.

PR #430, opened 2021-07-31
fix a few inference issues

ChrisRackauckas:


KristofferC:
Any reason why RandomNumbers.jl had to be restricted?

ChrisRackauckas:
It's seeding was broken on two versions, and I'm still not sure it isn't broken on v1.5.2. I'll need to do a deep check of the test errors to see if they are stochastic or distributional.

Issue #435, opened 2021-09-17
seed given by solver keyword argument does not work for sparse noise_rate_prototype


sho7014:
Hello, 

`seed` argument in `solve()` does not seem to work when `noise_rate_prototype` is specified using sparse arrays. 
Instead. `seed` can be appropriately set using `Random.seed!()`. 

You can see this behavior using the following code. 

```
using StochasticDiffEq, PyPlot, SparseArrays, Random

function f!(du,u,p,t)
    du[1] = 0.0
end

function g!(du,u,p,t)
    du[1,1] = 1.0
    du[1,3] = 1.0
end

dt = 1e-2
u0 = [0.0]
tspan = (0.0,0.1)

#seed does not work
sde_prob = SDEProblem(f!, g!, u0, tspan, noise_rate_prototype=sparse([1.0 0.0 1.0]))
sol = solve(sde_prob, EulerHeun(), dt=dt, seed=1)

#seed works
sde_prob = SDEProblem(f!, g!, u0, tspan, noise_rate_prototype=[1.0 0.0 1.0])
sol = solve(sde_prob, EulerHeun(), dt=dt, seed=1)

#seed works
sde_prob = SDEProblem(f!, g!, u0, tspan, noise_rate_prototype=sparse([1.0 0.0 1.0]))
Random.seed!(1)
sol = solve(sde_prob, EulerHeun(), dt=dt)
```

PR #436, opened 2021-09-26
Keep noise vectors dense

ChrisRackauckas:
Previously we would create the noise vector via `noise_rate_prototype[1,:]`, which is the right size, but if the matrix is sparse that would be a sparse vector. This causes two issues. One is performance. The noise vector is always dense (since otherwise you would just make it smaller in the non-diagonal case), so using a sparse vector is just a bad idea. But secondly, it turns out that filling a sparse random vector that starts with 0 memory can lead to some undefined behavior, which causes the first random numbers to be truly random. Thus this performance fix is also fixes https://github.com/SciML/StochasticDiffEq.jl/issues/435, which I would thus say is the best of all worlds.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
 version = "6.37.1"
 
 [deps]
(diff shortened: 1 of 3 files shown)


sho7014:
Thank you for the clear explanation and the fix.

PR #440, opened 2021-10-08
Try bumping RandomNumbers.jl

ChrisRackauckas:


frankschae:
Hmm... I'll take a look. IIRC most weak solvers are using about 1e9 trajectories for the convergence tests in the respective papers. So, to make them pass for a smaller number of trajectories, it was necessary to tweak them a bit.

ChrisRackauckas:
A tolerance bump may be required on those stragglers. As long as we know they do converge correctly at 1e9

frankschae:
yeah maybe I'll double-check 1e9 trajectories outside of the CI.

ChrisRackauckas:
https://github.com/SciML/StochasticDiffEq.jl/pull/440/checks?check_run_id=3914104494#step:6:363 looks like it is converging, just really really slowly. I'm going to merge since not allowing RandomNumbers.jl is giving the package manager some issues, but this still needs to get sorted out.

PR #443, opened 2021-11-16
Move to `juliaecosystem`

staticfloat:
We don't have GPU runners on the sandbox runners yet, but we may eventually get them. Until then, we can continue to use `juliagpu`.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -6,7 +6,7 @@ steps:
       - JuliaCI/julia-test#v1:
           coverage: false
     agents:
-      queue: "juliacpu"
+      queue: "juliaecosystem"
     env:
       GROUP: 'OOPWeakConvergence'
     timeout_in_minutes: 240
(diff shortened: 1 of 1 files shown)


staticfloat:
Hmmm, not sure what could be the reason, other than that this is a shared system, but looking into it, it doesn't seem to be overloaded.  Are these benchmarks supposed to be highly parallel?  It's possible that previously they were given all 128 processors to slam, and now they're only being given 16?

ChrisRackauckas:
Yeah, those benchmarks are rather large for a very good reason. Can we not get the machine back? We got it for this purpose, so we should reprovision so we can do this again. 😅

staticfloat:
@ChrisRackauckas actually, looking at the test history here, it looks like this is not the issue at all.  The `OOPWeakConvergence` and `IIPWeakConvergence` tests have been failing for a long time, even when they were running on an exclusive machine.

staticfloat:
And watching `htop` while the tests are running, it appears that things are running single-threaded.  Is that expected?

ChrisRackauckas:
Oh @frankschae this needs to get checked still.

frankschae:
hmm, I'll have a look -- I expected the tests to be multi-threaded, because they should use `EnsembleThreads()`, see https://github.com/SciML/DiffEqDevTools.jl/blob/37fde034a03b8dcf440613a81df1483a06ea25f9/src/convergence.jl#L43 .

PR #444, opened 2021-11-20
Enable multithreaded Julia on large tests

staticfloat:
We need to explicitly tell Julia (when running the tests) to run with as many threads as the machine provides.


staticfloat:
`htop` shows this as quite effective:

image

ChrisRackauckas:
image

PR #445, opened 2021-12-02
CI for LTS

ChrisRackauckas:


ChrisRackauckas:
@frankschae this is the last one I've been working on.

ChrisRackauckas:
We might need to move some tests to the buildkite, but let's see if this is ready to merge first 😅

ChrisRackauckas:
Some of those test failures on v1.7 look way off?

frankschae:
Yes... but I think the errors have already alternated at least once (i.e. once it failed at 1.6, then at 1.7) when increasing the number of trajectories. 
The number of trajectories is still on the lower end. Roessler uses 1e8 to 1e9 (https://link.springer.com/article/10.1007/s10543-007-0130-3)

ChrisRackauckas:
I'd say, swap all of those big weakconvergence tests to buildkite right now. This might take awhile longer than I had originally thought 😅

frankschae:
The tests, where I've just increased the number of trajectories, were still relatively small up to now (except the 1e7 ... that one is indeed a little odd ..). Basically,  we haven't seen those before because weak_convergence2 was failing and the other tests were canceled and didn't run through.

frankschae:
> I'd say, swap all of those big weakconvergence tests to buildkite right now. This might take awhile longer than I had originally thought 😅

yup, good idea. I'll do it.

ChrisRackauckas:
oops, looks like this is done but a regression on master needs to be fixed. Good time to have the tests working. Thanks @frankschae  😅

PR #448, opened 2021-12-20
Optimize CI choices a bit more

ChrisRackauckas:
I think those ones don't need the big exclusive node. @frankschae any others?


frankschae:
I think OOPWeakConvergence and IIPWeakConvergence could probably be used without the `exclusive` flag (because they hold the weak convergence tests for the strong solvers, so by and large they require fewer trajectories).

PR #449, opened 2021-12-20
Remove exclusive on Buildkite tests

ChrisRackauckas:


ChrisRackauckas:
@frankschae looks like the machines crash if those ones are removed. All because of SROCKC2. I wonder if that specific test should just be moved to a separate section.

frankschae:
Perhaps, it should. I would guess that it could make a slight difference because if we separate SROCKC2, only one test group has to wait for the exclusive workers. While two test groups are affected currently.

Issue #458, opened 2022-01-29
Integrate LevyArea.jl into StochasticDiffEq.jl


ChrisRackauckas:
https://github.com/stochastics-uni-luebeck/LevyArea.jl
https://arxiv.org/abs/2201.08424

This paper and library is great @fkastner! As noted in the paper, LevyArea.jl has new and faster implementations of the Levy area calculations. We might as well use them in our higher order integrators. @frankschae could you take a look at how difficult that would be? I assume it wouldn't be too difficult as it should be a direct stand-in for `StochasticDiffEq.get_iterated_I!(h,W,nothing,iip,p) `. Note that we should then update the citation materials in each of the higher order algorithms which use it to point to their paper.

(P.S. Note that you're citing most of the Julia libraries incorrectly. Julia has a standard of using Citation.bib files, so it's recommended you cite the associated papers for the library. Here for example it's: https://github.com/SciML/StochasticDiffEq.jl/blob/master/CITATION.bib. https://github.com/SebastianM-C/PkgCite.jl helps automate this.)

fkastner:
Hi Chris,
thanks for your interest in our package.

The integration should probably look something like this:
```julia
using LevyArea
import StochasticDiffEq: get_iterated_I

function get_iterated_I(dt, dW, ::Nothing, alg::LevyArea.AbstractIteratedIntegralAlgorithm, p=nothing, c=1, γ=1//1)
    if isnothing(p)
        ε = c*dt^(γ+1//2)
        p = terms_needed(length(dW), dt, ε, alg, MaxL2())
    end
    I = LevyArea.levyarea(dW/√dt, p, alg)
    I .= 0.5.*dW.*dW' .+ dt.*I
end
```
and then you need some way to choose between the different algorithms. 

For a simple test I just overwrote `get_WikJ`:
```julia
using StochasticDiffEq

function StochasticDiffEq.get_WikJ(ΔW,prob,alg::RKMilGeneral)
    return MronRoe()
end

u₀=[0.0,0.0,0.0]
f(u,p,t) = [0.0,0.0,0.0]
g(u,p,t) = [1.0 0.0;0.0 1.0;0.0 u[1]]
dt = 1//2^10
tspan = (0.0,1.0)
prob = SDEProblem(f,g,u₀,(0.0,1.0); noise_rate_prototype=zeros(3,2))
sol = solve(prob, RKMilGeneral(), dt=dt, adaptive=false);
```

[...]

PR #459, opened 2022-02-07
Switch to LevyArea.jl

frankschae:
See #458  
https://github.com/stochastics-uni-luebeck/LevyArea.jl
https://arxiv.org/abs/2201.08424

To do:
- [x] Need to check AD compatibility [ see https://github.com/stochastics-uni-luebeck/LevyArea.jl/issues/6 ]
- [x] Update defaults [defaults to using LevyArea, allows for selecting specific routine, still contains separate implementations for diagonal and commutative noise which probably should be handled in a follow-up PR.]
- [x] Update tests [no path-wise convergence tests]
- [x] Update citations [ [docs](https://diffeq.sciml.ai/dev/solvers/sde_solve/#Iterated-Integral-Approximations) should be updated once this is merged.]


test/levy_areas.jl
```diff
@@ -84,7 +84,7 @@ Problem 2.3.3 from
 Kloeden, P. E., Platen, E., & Schurz, H. Numerical solution of SDE through computer
 experiments. Springer Science & Business Media. (2012)
 """
-function test_path_convergence(Wik, dt = 1.0, ps = [Int(1e1),Int(1e3),Int(1e6)])  
+function test_path_convergence(Wik, dt = 1.0, ps = [Int(2e1),Int(1e3),Int(1e6)])
```
fkastner, line 87: This doesn't work because of the order in which the random numbers are generated. It is *not* the case that two different values of `p` share the first few random numbers in the same order.

frankschae: Next, I planned to average over some seeds and see if this eliminates the errors that sometimes `p=1e3` could be worse than `p=1e2`.

fkastner: > I agree that this is not generally the case. However, shouldn't this work once I fix the seed? So, I force that the larger `p` values contain the random numbers of smaller `p` values
> 
> ```
> Random.seed!(seed)
> I = StochasticDiffEq.get_iterated_I(dt, W.dW, W.dZ, Wik, p, 1)
> ```

That's exactly what doesn't work. There are different types of random numbers needed in the algorithm, lets call them for simplicity `a_1` to `a_p`, `b_1` to `b_p` and `c`.
If `p` is 4 and our random number generator (again for simplicity) generates increasing integers, we get:
`a = 1 2 3 4`
`b = 5 6 7 8`
and `c = 9`.
But for `p` equal to 5 we get:
`a = 1 2 3 4 5`
`b = 6 7 8 9 10`
and `c = 11`.
Thus they don't share the values because of the order in which they are generated/needed.

src/iterated_integrals.jl
```diff
@@ -5,141 +5,11 @@ by the J.
...
-  #@show WikA, sqrt(dt*ρp)*(kronprod1-kronprod2) , WikA + sqrt(dt*ρp)*(kronprod1-kronprod2)
-  #@show sqrt(dt*ρp)
-  @.. WikJ = WikJ/2 + WikA + sqrt(dt*ρp)*(kronprod1-kronprod2) # Ito: -UniformScaling(dt))/2
-  return nothing
-end
-
-
-
-# Wiktorssson Approximation
+# Wiktorssson Approximation (for diagonal and commutative noise processes where the Levy area approximation is not required.)
```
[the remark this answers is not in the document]
frankschae: Yup, agreed. The names go back to the first attempt where the goal was to implement Wiktorsson's algorithm (plus handling the special cases efficiently where the Levy area approximation is not required). I already updated the one that's user-facing
`IIWiktorsson` -> `IILevyArea` (open for better suggestions and suggestions for the other ones).

> Also I'm not really sure why you have different routines for diagonal and commutative noise? They should be the same, but there are some slight differences in your implementation I don't understand.

This is because we represent the diffusion matrix for diagonal noise by a vector, see 
https://diffeq.sciml.ai/dev/tutorials/sde_example/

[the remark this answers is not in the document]
fkastner: > This is because we represent the diffusion matrix for diagonal noise by a vector, see
> [diffeq.sciml.ai/dev/tutorials/sde_example](https://diffeq.sciml.ai/dev/tutorials/sde_example/)

Wouldn't it be easier to just wrap the return value of g in a `Diagonal`? That way the intention is clearer and the solvers don't need to be special cased.
But ok, that was probably discussed when the interface was created and is set in stone now.

ChrisRackauckas:
For strong convergence (and adaptivity) we would have to come up with a scheme so that we could easily pullback the Brownian paths. In the diagonal and commutative cases this is easy because it is done with the Brownian bridge, and the diagonal 1.5 case can be handled with a psuedo-Brownian process because the iterated integral approximation turns out to just be essentially two Brownian processes. I don't know what would have to happen with the non-commutative case though since it takes `n` random numbers?

frankschae:
The test failures (StochasticDelayDiffEq/SROCKC2) look unrelated to me.

Project.toml
```diff
@@ -14,6 +14,7 @@ DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 FiniteDiff = "6a86dc24-6348-571c-b903-95158fe2bd41"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+LevyArea = "2d8b4e74-eb68-11e8-0fb9-d5eb67b50637"
```
devmotion, line 17: This needs a compat bound, otherwise you won't be able to add a new release to the registry.

fkastner:
Is the distinction between in-place and out-of-place still necessary for the iterated integrals? Especially since LevyArea.jl currently doesn't provide that distinction either. There is no reason why the in-place-ness of the iterated integral computations has to be coupled to the in-place-ness of the problem itself, or am I missing something here?

frankschae:
I would say we could either remove the inplaceness completely, i.e., by removing it also from all SDE solvers 
https://github.com/SciML/StochasticDiffEq.jl/blob/c802f6ec6a8cb06e0d5797fc7605d7cb85fb9e25/src/perform_step/low_order.jl#L449
(For problems with mutating functions, the iip formulations were originally used to avoid allocations -- this doesn't work for the methods from LevyArea.jl.) 
Alternatively, we can use this wrapper that offers no computational advantage but allows us to remove allocations for the other cases (`IICommutative()`) in principle.

ChrisRackauckas:
The SDDE test is just noisy. I need to fix that when I get back.

ChrisRackauckas:
Continuing the discussion on fully inplace in https://github.com/stochastics-uni-luebeck/LevyArea.jl/issues/7 . While it could hit performance, for now we can probably just skip it.

fkastner:
> Alternatively, we can use this wrapper that offers no computational advantage but allows us to remove allocations for the other cases (`IICommutative()`) in principle.

Yes, that was my suggestion. Basically remove the out-of-place functions. It should be possible to use in-place Levy area approximations (no matter how much of it is in-place) even for out-of-place problems, shouldn't it?

ChrisRackauckas:
out of place would still be required for some forms of automatic differentiation and for use cases with immutable arrays and static arrays. So that still would have a place in the grant scheme of things.

PR #477, opened 2022-06-24
test master

ChrisRackauckas:


ChrisRackauckas:
Is that a weak convergence 4 timing regression @frankschae ?

frankschae:
yeah looks like... it used to run in ~4h, see, e.g., https://github.com/SciML/StochasticDiffEq.jl/pull/459. Also, the test failures in weakadaptivegpu look real?

ChrisRackauckas:
Yes

Issue #481, opened 2022-07-14
UndefVarError: J not defined from line 60 of implicit_split_step.jl


evolbio:
From solve ```(prob, S.solver, u0=u_init, p=p[S.ddim+1:end])```
for prob as SDEProblem with no special args, and solver as ISSEM(). 
```
ERROR: UndefVarError: J not defined
Stacktrace:
[...]

PR #483, opened 2022-07-19
adaptive stepping `ISSEM` fixes

frankschae:
Fixes the original issue https://github.com/SciML/StochasticDiffEq.jl/issues/481:
```
ERROR: UndefVarError: J not defined
```
for oop `SDEProblem`s. 

In addition, there were some inconsistencies in the adaptive sub-routine. 
for oop problems:
- There was a `dt` missing from the translation of https://github.com/SciML/StochasticDiffEq.jl/blob/b9806da512d1897cc5e8f9af37f4f15968d1f4cc/src/perform_step/lamba.jl#L23 to the Jacobian version.

for in-place problems:
- `tmp` (and some other variables) were changed before the use in the adaptive routine. Therefore, the stepping behavior did not agree with the oop version or the (similar) one from `LambdaEM`. I have now moved this part up. The variables `k` and `dz` have to be cached additionally in the alg cache. Otherwise, they are overwritten in the 
  ```julia
  z = OrdinaryDiffEq.nlsolve!(nlsolver, integrator, cache, repeat_step)  
  ```
  call.
- `gtmp2` and `gtmp` were identical because of the missing `copy` in the cache

### DIFF
--- a/src/caches/implicit_split_step_caches.jl
+++ b/src/caches/implicit_split_step_caches.jl
@@ -6,33 +6,38 @@
   gtmp2::rateType
   nlsolver::N
   dW_cache::randType
+  k::uType
+  dz::uType
 end
 
-function alg_cache(alg::ISSEM,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,jump_rate_prototype,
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):
Awesome, I wonder if Torkel can take a crack at trying this improved version.

PR #484, opened 2022-07-20
Remove unnecessary overload

YingboMa:
Needs https://github.com/SciML/OrdinaryDiffEq.jl/pull/1698

@ChrisRackauckas how should we release this? OrdinaryDiffEq's downstream test won't pass because of this, while this alone would be broken.


ChrisRackauckas:
Merge OrdinaryDiffEq.jl first and just follow up with a patch here.

PR #489, opened 2022-07-29
SDE reversal for scalar noise processes

frankschae:
Tackle open issues regarding time reversion of Ito SDEs from: 

https://github.com/SciML/SciMLSensitivity.jl/issues/279
https://github.com/SciML/DiffEqNoiseProcess.jl/pull/62

@ChrisRackauckas maybe we should move the tests from DiffEqNoiseProcess.jl here? (It seems that it is worth to test several algorithms and not only if NoiseGrid, NoiseWrapper etc. work).


ChrisRackauckas:
Possibly?

PR #494, opened 2022-08-17
generalize reset to work with NoiseFunction

### LINKED ISSUE #493: `NoiseFunction` not resetting for new solves
For example. With
```julia
using StochasticDiffEq, DiffEqNoiseProcess

f(u, p, t, Y) = Y * u

y(a, p, t) = -t
Y = NoiseFunction(0.0, y, reset=true)

u0 = 1.0
tspan = (0.0, 1.0)

prob = RODEProblem(f, u0, tspan, noise=Y)
```
I get
```
julia> sol = solve(prob, RandomEM(), dt=1/100)
retcode: Success
Interpolation: 1st order linear
t: 101-element Vector{Float64}:
 0.0
 0.01
 ⋮
 1.0
u: 101-element Vector{Float64}:
 1.0
 1.0
 ⋮
 0.6085659649572785

julia> sol = solve(prob, RandomEM(), dt=1/100)
retcode: Success
Interpolation: 1st order linear
t: 101-element Vector{Float64}:
 0.0
 0.01
 0.02
 ⋮
 0.9800000000000006
 0.9900000000000007
 1.0
u: 101-element Vector{Float64}:
 1.0
 0.99
 0.980001
 ⋮
 0.23070445114140511
 0.22613650300880528
 0.22163638659893034
```
Notice `u` keeps decreasing (the decay rate `Y` does not reset for the new solves and keeps getting faster). After the third run, `Y.curt` is near 3.0:
```julia
julia> Y.curt
2.9999999999999782
```

The problem is that `solve` expects the noise to have the field `t`, which is not the case for a `NoiseFunction`: https://github.com/SciML/StochasticDiffEq.jl/blob/master/src/solve.jl#L416

rmsrosa:
This changes the way `reset` is done for a new `solve` in a way to accommodate noises of type `NoiseFunction`.

It checks for `W.curt != t` instead of `W.t[end] != t`, so it works for the usual noises, which do have the field `t` and for which `W.curt` is the same as `W.t[end]`, and for `NoiseFunction`s, which do not have the field `t`. 

To be fully functional, this depends on implementing a proper dispatch of `reinit!` to `NoiseFunction` in [SciML/DiffEqNoiseProcesses.jl](https://github.com/SciML/DiffEqNoiseProcess.jl), as proposed in the PR https://github.com/SciML/DiffEqNoiseProcess.jl/pull/114.

This should close #493.

With these things implemented, we get the proper behavior (*c.f.* #493):
```julia
using StochasticDiffEq, DiffEqNoiseProcess

f(u, p, t, Y) = Y * u

y(a, p, t) = -t
Y = NoiseFunction(0.0, y, reset=true)

u0 = 1.0
tspan = (0.0, 1.0)

prob = RODEProblem(f, u0, tspan, noise=Y)
```

```julia
julia> sol = solve(prob, RandomEM(), dt=1/100)
retcode: Success
Interpolation: 1st order linear
t: 101-element Vector{Float64}:
 0.0
 ⋮
 1.0
u: 101-element Vector{Float64}:
 1.0
 ⋮
 0.6085659649572785

julia> sol = solve(prob, RandomEM(), dt=1/100)
retcode: Success
Interpolation: 1st order linear
t: 101-element Vector{Float64}:
 0.0
 ⋮
 1.0
u: 101-element Vector{Float64}:
 1.0
 ⋮
 0.6024803053077056

julia> Y.curt
1.0
```

And preserving the behavior for "regular" noises

```julia
julia> prob = RODEProblem(f, u0, tspan, noise=W)

RODEProblem with uType Float64 and tType Float64. In-place: false
timespan: (0.0, 1.0)
u0: 1.0

julia> sol = solve(prob, RandomEM(), dt=1/100);

julia> sol = solve(prob, RandomEM(), dt=1/100);

julia> W.curt
1.0
```


ChrisRackauckas:
@oscarsmith https://github.com/SciML/StochasticDiffEq.jl/runs/7886061441?check_suite_focus=true#step:6:424 I think you caused a downstream failure.

ChrisRackauckas:
@frankschae can you track down the regression in the weak adaptive GPU? I think it's a KernelAbstractions.jl Rational support bug, so @vchuravy might need an MWE.

But this looks fine to merge.

ChrisRackauckas:
This needs a test which would catch it.

rmsrosa:
Yeah, a test is good. Just added some.

PR #497, opened 2022-08-20
Move to SDEProblemLibrary

ChrisRackauckas:


Project.toml
```diff
@@ -58,14 +58,14 @@ julia = "1.6"
 DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
 DiffEqDevTools = "f3b72e0c-5b89-59e1-b016-84e28bfd966d"
 DiffEqOperators = "9fdde737-9c7f-55bf-ade8-46b3f136cc48"
-DiffEqProblemLibrary = "a077e3f3-b75c-5d7f-a0c6-6bc4c8ec64a9"
 LinearSolve = "7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
+SDEProblemLibrary = "c72e72a9-a271-4b2b-8966-303ed956772e"
```
devmotion, line 65: Might be useful to add compat entries for test dependencies to avoid that breaking releases are picked up automatically and break the tests.

ChrisRackauckas: I thought I checked before and the package manager ignores them (even though it parses correctly)

devmotion: I was quite certain that it is respected. The test dependencies are bounded in this way in e.g. ChainRulesCore: https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/Project.toml

PR #507, opened 2022-09-30
WIP: fix SROCK2 for NoiseWrapper noise

### LINKED ISSUE #505: SROCK2() not working with NoiseWrapper
SROCK2() is not working with a `W::NoiseWrapper` noise as that has no `W.rng` field, conflicting with

https://github.com/SciML/StochasticDiffEq.jl/blob/5c13cc7f677eaf231b7e4521595f26aea612c250/src/perform_step/SROCK_perform_step.jl#L335

axsk:
fixes #505


axsk:
Looking at the other uses of `gen_prob` I just realized that some methods don't care about `W.rng` and just use `rand()` anyways.

ChrisRackauckas:
What's the status of this?

axsk:
Sorry for leaving this hanging, was on vacation and then forgot about it.
I think it is done.

The one thing I am not sure of (looking at it now) is that I am not using the `oftype(W.dW, ...` any more.
I don't know if that poses any problems.

ChrisRackauckas:
Add a test?

axsk:
Well the canonical tests seems to be running. I think the `oftype` only comes to play when `W.dW` is of different type then `rand(W.rng)`, and I wouldn't know when this is the case.

axsk:
Or do you mean a test that the `NoiseWrapper` is actually respected?

ChrisRackauckas:
>  Or do you mean a test that the NoiseWrapper is actually respected?

That. Try solving it with a `NoiseWrapper` and seeing if it succeeds / fails. Also try other noise types like NoiseGrid`

axsk:
Could you hint me at where to put these tests?

ChrisRackauckas:
https://github.com/SciML/StochasticDiffEq.jl/blob/master/test/noise_type_test.jl would probably be good

ChrisRackauckas:
Closing as part of StochasticDiffEq.jl archival. SDE solvers have been migrated to OrdinaryDiffEq.jl (see https://github.com/SciML/OrdinaryDiffEq.jl/pull/3154). The SROCK2 NoiseWrapper fix (related to #505) should be verified/re-done in the new codebase.

Issue #519, opened 2023-01-05
Functional form for BAOAB solver


jebej:
From PR #397, it is stated that the functional form assumed for the SDE with the `BAOAB` solver is
```
dq = v*dt
dv = f1*dt - gamma*v*dt + g*sqrt(2gamma)*dW
```
This is not what I would expect, since it has an extra `-gamma*v*dt` term. Is there a reason for this difference?

Comparing with the ODE case in the following script, the solutions are different:

```julia
using OrdinaryDiffEq, StochasticDiffEq, PyPlot

jj!(ϕ′,ϕ,τ,β,I) = (I - ϕ′ - sin(ϕ))/β
jj!(ϕ′,ϕ,p::Tuple,τ) = jj!(ϕ′,ϕ,τ,p[1],p[2])
jj!(ϕ′::Vector,ϕ::Vector,p::Tuple,τ) = jj!(ϕ′[],ϕ[],τ,p[1],p[2]) # special version to workaround #517 (fixed on master)
identity_f(v, u, p, t) = v # needed to form second order dynamical ODE

setup_jj(β,I,ϕ_init=0.0,ϕ′_init=0.0,tspan=(0.0,50.0)) = DynamicalODEProblem{false}(jj!,identity_f,ϕ′_init,ϕ_init,tspan,(β,I))

g(u,p,t) = 0.01 # small noise
setup_noisy_jj(β,I,ϕ_init=[0.0],ϕ′_init=[0.0],tspan=(0.0,50.0)) = DynamicalSDEProblem{false}(jj!,identity_f,g,ϕ′_init,ϕ_init,tspan,(β,I))

function analyze_junction_dc(res, clear=true)
    noisy = res isa RODESolution
    V = getindex.(res.u,1)
    I = sin.(getindex.(res.u,2))
    clear && clf()
    plot(res.t, V, label="\$V/φ₀\$"*ifelse(noisy," (Langevin)"," (noise-free)"))
    plot(res.t, I, label="\$I/I_c\$"*ifelse(noisy," (Langevin)"," (noise-free)"))
    xlabel("Time τ")
    legend()
    grid(true)
    display(gcf())
end

prob = setup_jj(1.25,2.0)
res = solve(prob,VelocityVerlet(),dt=1/100)
analyze_junction_dc(res)

prob = setup_noisy_jj(1.25,2.0)
res = solve(prob,BAOAB(1),dt=1/100)
analyze_junction_dc(res, false)
```

[...]

PR #520, opened 2023-01-06
remove gamma param from noise term for BAOAB alg

jebej:
As discussed in #519, I want to experiment with removing the `gamma` dependence attached to the noise term for the `BAOAB` algorithm.

The main issue is that if you do want the `gamma` dependence, you also need a dependence on `dt` (coming from the `c1` variable in the original code). While this `dt` dependence could be user-coded in the `g` noise function, it's not an ideal interface.

Suggestions welcome.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -866,5 +866,6 @@ dv = f(v,u) dt - \gamma v dt + g(u) \sqrt{2\gamma} dW
 """
 struct BAOAB{T} <: StochasticDiffEqAlgorithm
   gamma::T
+  scale_noise::Bool
 end
-BAOAB(;gamma=1.0) = BAOAB(gamma)
+BAOAB(;gamma=1.0, scale_noise=true) = BAOAB(gamma, scale_noise)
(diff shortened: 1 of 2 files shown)


rmsrosa:
I am not familiar with the method, but browsing the discussions, what about, instead of removing `c2`, adding another keyword argument to `BAOAB()` with the option of changing the default behavior?

ChrisRackauckas:
I just suggested that same thing in another way. I agree it should just have two kwargs.

jebej:
If we want another keyword argument, the only thing I can think of is something like `scale_noise=true`, which keeps the current behavior, and when false just sets the `c2` variable to 1. Anything else can be handled in the user function.

ChrisRackauckas:
Or if it's another float, you just make the default `c2 = gamma`.

jebej:
If we want the original behavior from the paper we need `dt`:
```
c1 = exp(-alg.gamma*dt)
c2 = sqrt(1 - c1^2)
```

ChrisRackauckas:
It's not already scaling by dt?

PR #522, opened 2023-03-03
import AbstractController from OrdinaryDiffEq.jl

frankschae:
If Controllers should subtype `StochasticDiffEq.AbstractController` this will allow us to implement custom adaptive schemes easily.

```julia
struct CustomController <: StochasticDiffEq.AbstractController
end

function StochasticDiffEq.stepsize_controller!(integrator::StochasticDiffEq.SDEIntegrator, controller::CustomController, alg)
    u = integrator.u
    integrator.q = max(...)
    nothing
end

function StochasticDiffEq.step_accept_controller!(integrator::StochasticDiffEq.SDEIntegrator, controller::CustomController, alg)
    integrator.dtnew = 1 / integrator.q * oneunit(integrator.dt)
    nothing
end

sol = solve(prob, EM(), dt=dt, adaptive=true, controller=CustomController())
```

### DIFF
--- a/src/StochasticDiffEq.jl
+++ b/src/StochasticDiffEq.jl
@@ -14,7 +14,7 @@ using DocStringExtensions
          beta2_default, beta1_default, gamma_default,
          qmin_default, qmax_default, qsteady_min_default, qsteady_max_default,
          stepsize_controller!, accept_step_controller, step_accept_controller!,
-         step_reject_controller!, PIController, DummyController
+         step_reject_controller!, PIController, DummyController, AbstractController
 
   using UnPack, RecursiveArrayTools, DataStructures
   using DiffEqNoiseProcess, Random, ArrayInterface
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Maybe this should be implemented in OrdinaryDiffEq and then just used with any solver?

frankschae:
> Maybe this should be implemented in OrdinaryDiffEq and then just used with any solver?

Can you elaborate? I thought that it's fine to let a user define their own controller type as a subtype of `AbstractController` and then dispatch on it to change the adaptive scheme. `AbstractController` is defined in OrdinaryDiffEq and only imported here. I don't know how to implement a generically valid `integrator.q = max(...)`.

ChrisRackauckas:
I mean, why not implement CustomController in OrdinaryDiffEq and then document it as a feature for ODEs, SDEs, DDEs, DAEs, and SDDEs?

frankschae:
What would that look like? Would we only add
```julia
struct CustomController <: StochasticDiffEq.AbstractController
end
```
to OrdinaryDiffEq? Wouldn't that be worse than leaving the type "free"? I have in mind that users might want to do something sophisticated and it would be good if `CustomController` could hold additional fields (but adding it to the docs would be great of course).

ChrisRackauckas:
Yeah, I think some documented type where a user gives some functions like that.

frankschae:
Isn't `AbstractController` basically already that type (so you can subtype it with a concrete type with its own fields and do all the fancy things you want to do)?

ChrisRackauckas:
Documenting it might be all that's necessary.

frankschae:
Maybe on this page https://docs.sciml.ai/DiffEqDocs/stable/extras/timestepping/?

ChrisRackauckas:
That seems right.

PR #527, opened 2023-03-16
destats -> stats

ChrisRackauckas:


YingboMa (approved):

SebastianM-C:
@ChrisRackauckas I think this needs a compat bound on DiffEqBase, as it errors with DiffEqBase versions before v6.122.0

PR #531, opened 2023-05-16
overload _alg_autodiff

vpuri3:
fix downstream errors in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1917

@gaurav-arya can you review?


ChrisRackauckas:
To make the transition simpler, it may be good to yank all other v2 releases.

xtalax:
@ChrisRackauckas  v2 of what, and which packages?

ChrisRackauckas:
yank releases of StochasticDiffEq which allow SparseDiffTools v2.

ChrisRackauckas:
Looking around, we never did do a SparseDiffTools v2 compatible release so the versions bounds naturally work well here.

vpuri3:
it's grabbing ODE.jl 6.51.2. 

https://github.com/SciML/StochasticDiffEq.jl/actions/runs/5095012736/jobs/9159498470?pr=531#step:6:60

vpuri3:
@ChrisRackauckas looks like this needs an OrdinaryDiffEq release.

https://github.com/SciML/StochasticDiffEq.jl/actions/runs/5098413712/jobs/9165536977?pr=531#step:5:20

PR #535, opened 2023-06-24
Update the last few retcode symbols to ReturnCode

danielwe:
Looks like #510 missed these.


ChrisRackauckas (approved):

danielwe:
When do you think you'll cut a new release? The current version (without this PR) emits a deprecation warning when using `terminate!` in a callback, and it gets rather noisy in large ensemble runs.

Issue #536, opened 2023-07-07
Overwriting keyword `adaptive`


Jonas-a-Zimmermann:
Currently `solve` lets you set the keyword `adaptive` to decide whetever the solver should use adaptive or fixed time step methods.

I think this can, for non adapative solvers, lead to undefined behavior?
Either way, i think a more helpfull error message might help the user troubleshooting.

Given a `SDEproblem prob`
 `solve(prob, EM())`
Throws the error: `ERROR: Fixed timestep methods require a choice of dt or choosing the tstops`

However if I write
`solve(prob, EM(); adaptive=true)`
shows: `┌ Warning: dt(9.094947017729282e-13) <= dtmin(9.094947017729282e-13) at t=1.3398645121662485e-8, and step error estimate = 1.0. Aborting. There is either an error in your model specification or the true solution is unstable.`

PR #537, opened 2023-07-07
added Error on adaptive for non adaptive algorithms

Jonas-a-Zimmermann:
Implements a more clear error message for #536 .


Jonas-a-Zimmermann:
`solve(problem, EM(); adaptive=true, dt=0.01)`
now leads to 
`ERROR: The given solver is a Fixed timestep method and does not support adaptivity.`

ChrisRackauckas:
Thanks! Sorry I missed the email message on this, but at least it's in now.

PR #545, opened 2023-09-11
Dynamical integrators

HadrienNU:
Following discussion in #259 this is a proposed implementation of the ABOBA and OBABO integrators (GJ set of integrators coming later).  I have also modified the BAOAB implementation to allow matrix as friction (the gamma parameter) in the equation.

That seems to work, but I am not sure to have done everything properly (in particular to verify is gamma is a matrix or not). So let me know if there are changes to do.

Also, I would like to add the integrator to the documentation, but I am not sure of where I should propose changes.


src/solve.jl
```diff
@@ -210,10 +210,13 @@ function DiffEqBase.__init(
...
   rateType = typeof(rate_prototype) ## Can be different if united
 
-  if prob.f isa DynamicalSDEFunction
-    noise_rate_prototype = rate_prototype.x[1]
-  elseif is_diagonal_noise(prob)
-    noise_rate_prototype = rate_prototype
+
+  if is_diagonal_noise(prob)
+    if prob.f isa DynamicalSDEFunction
+      noise_rate_prototype = rate_prototype.x[1]
```
ChrisRackauckas, line 216: That seems like an odd way to put it. In other words, it's diagonal on one part of the system?

ChrisRackauckas: I wonder if we just shouldn't classify that as diagonal noise at all, but it looks like it doesn't lead to too many required changes.

ChrisRackauckas:
I think the implementation looks reasonable. @frankschae could you take a look through it as well?

src/solve.jl
HadrienNU: But the proposed change here is to allow DynamicalSDEFunction to have non diagonal noise, as the previous implementation assume the noise was diagonal.

test/sde/sde_dynamical.jl
```diff
@@ -19,6 +19,15 @@ g(u,p,t) = 1
     sim1  = analyticless_test_convergence(dts,prob1,BAOAB(gamma=γ),(1/2)^10;trajectories=Int(2e2),use_noise_grid=false)
     display(sim1.𝒪est)
     @test abs(sim1.𝒪est[:weak_final]-1) < 0.3
+
+    sim1  = analyticless_test_convergence(dts,prob1,ABOBA(gamma=γ),(1/2)^10;trajectories=Int(5e2),use_noise_grid=false)
+    display(sim1.𝒪est)
+    @test abs(sim1.𝒪est[:weak_final]-1) < 0.3
```
frankschae, line 25: Weak final order 1 seems strange to me -- aren't these methods supposed to be second-order in weak convergence? 

(Having said this, based on the paper https://arxiv.org/abs/1304.3269, it seems that the error may increase when you lower the step size (Figure 3) and second-order convergence can only be seen over very small intervals (Figure 4), so I am not too worried but maybe we should still try to reproduce the plots?)

src/algorithms.jl
```diff
@@ -869,3 +869,16 @@ struct BAOAB{T} <: StochasticDiffEqAlgorithm
   scale_noise::Bool
 end
 BAOAB(;gamma=1.0, scale_noise=true) = BAOAB(gamma, scale_noise)
+
+struct ABOBA{T} <: StochasticDiffEqAlgorithm
```
frankschae, line 873: maybe add a docstring that contains the SDE form:
Screen Shot 2023-09-19 at 11 44 02 AM

(or to the SDE solver docs https://docs.sciml.ai/DiffEqDocs/stable/solvers/sde_solve/ given that it doesn't match our common diagonal/non-diagonal noise classification exactly?)

test/sde/sde_dynamical.jl
HadrienNU: I agree that they should be 2nd order, but the numerics give me first order. I will try to reproduce the figure 4 in the paper over this smaller step size range to see what it give.

test/sde/sde_dynamical.jl
```diff
@@ -19,6 +19,15 @@ g(u,p,t) = 1
     sim1  = analyticless_test_convergence(dts,prob1,BAOAB(gamma=γ),(1/2)^10;trajectories=Int(2e2),use_noise_grid=false)
     display(sim1.𝒪est)
     @test abs(sim1.𝒪est[:weak_final]-1) < 0.3
+
+    sim1  = analyticless_test_convergence(dts,prob1,ABOBA(gamma=γ),(1/2)^10;trajectories=Int(5e2),use_noise_grid=false)
```
frankschae, line 23: convergence tests are failing on CI -- did you try increasing the number of trajectories? 500 seems relatively small.

test/sde/sde_dynamical.jl
frankschae: Do you have an update for this plot?

HadrienNU: fig3
Here is already the figure 3 of the paper (OBABO here correspond to BP there). Figure 4 is running, I should get it by tomorrow

HadrienNU: There is only 200 trajectories of 10^7 timesteps so less trajectories than the paper, but results are similar

HadrienNU: [fig4.pdf](https://github.com/SciML/StochasticDiffEq.jl/files/12854717/fig4.pdf)
And now the panel (a) and (c) of the figure 4 of the paper, I cannot go as far in stepsize since SciML detect instability at timestep starting from 0.24 and abort the computation so I went to smaller timestep to better see the scaling. There is less trajectories than in the paper, here 1000 of 10^7 timestep to be compared to 500 trajectories of 10^9 timestep, but the computation is quite demanding so I went for less timestep. The main features of the figure 4 are there (with BP=OBABO)

test/sde/sde_dynamical.jl
HadrienNU: I did increase the number of trajectories in tests and there are now passing (but longer computation time)

frankschae: Great! Maybe we should move the weak convergence tests also to the buildkite queue (@ChrisRackauckas?)

ChrisRackauckas: yes for sure.

src/algorithms.jl
ChrisRackauckas: Or maybe it needs a different problem type?

ChrisRackauckas:
I think the implementation looks pretty good to go? @frankschae what do you think should be done with the problem types here?

src/algorithms.jl
HadrienNU: There is already the DynamicalSDEProblem (was introduced by the previous implementation of BAOAB) that is supposed to be solved by those integrators

HadrienNU: Should I add some alg_compatible() functions to allow only DynamicalSDEProblem to be solved by Dynamical integrators?

ChrisRackauckas: yes that would be helpful in order to gate it to the right form.

HadrienNU: I just realize that the DynamicalSDEProblem constructor actually return a SDEProblem so alg_comptable functions won't be useful here

ChrisRackauckas: But it can dispatch on `prob.problem_type`

frankschae: yeah I think we should use `DynamicalSDEProblem` and it's a good idea to use `alg_compatible` to make sure that only BAOAB, ABOBA, etc. can be used to solve it.

test/sde/sde_dynamical.jl
```diff
@@ -3,22 +3,31 @@ Random.seed!(1)
...
+
+    sim1  = analyticless_test_convergence(dts,prob1,ABOBA(gamma=γ),(1/2)^10;trajectories=Int(1e5),use_noise_grid=false)
+    display(sim1.𝒪est)
+    @test abs(sim1.𝒪est[:weak_final]-2) < 0.5
+
+
+    sim1  = analyticless_test_convergence(dts,prob1,OBABO(gamma=γ),(1/2)^10;trajectories=Int(1e5),use_noise_grid=false)
     display(sim1.𝒪est)
-    @test abs(sim1.𝒪est[:weak_final]-1) < 0.3
+    @test abs(sim1.𝒪est[:weak_final]-1.5) < 0.5
```
ChrisRackauckas, line 30: This is ambiguous as to whether the weak is 1 or 2?

ChrisRackauckas:
I just realized I forgot about these. Let's see what's going on with those convergence tests and get this merged.

test/sde/sde_dynamical.jl
```diff
@@ -31,17 +40,116 @@ end
...
+    # @test abs(sim1.𝒪est[:weak_final]-2) < 0.3
+
+
+    sol1 = solve(prob1,OBABO(gamma=[γ,γ]);dt=1/10,save_noise=true)
+    prob2 = DynamicalSDEProblem(f1_harmonic_iip,f2_harmonic_iip,g_iip,v0,u0,(0.0,5.0); noise=NoiseWrapper(sol1.W))
+    sol2 = solve(prob2,OBABO(gamma=[γ,γ]);dt=1/10)
+
+    @test sol1[:] ≈ sol2[:]
+
+    # # Can't use NoiseGrid as noise is not generated with the correct size in convergence.jl. We require noise with shape of v.
```
frankschae, line 75: Can you not predefine a NoiseGrid with the correct size?

src/algorithms.jl
```diff
@@ -869,3 +869,16 @@ struct BAOAB{T} <: StochasticDiffEqAlgorithm
...
+  gamma::T
+  scale_noise::Bool
+end
+ABOBA(;gamma=1.0, scale_noise=true) = ABOBA(gamma, scale_noise)
+
+
+struct OBABO{T} <: StochasticDiffEqAlgorithm
```
frankschae, line 880: Can you add the same docstring as for BAOAB for ABOBA and OBABO?

frankschae (changes requested):
The `perform_step` functions of BAOAB, ABOBA, and OBABO look correct to me (except for `c1` and `c2` not being constants if `dt` is not constant).

src/caches/dynamical_caches.jl
```diff
@@ -1,42 +1,188 @@
...
   tmp::uTypeCombined
 end
 
 function alg_cache(alg::BAOAB,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,jump_rate_prototype,::Type{uEltypeNoUnits},::Type{uBottomEltypeNoUnits},::Type{tTypeNoUnits},uprev,f,t,dt,::Type{Val{false}}) where {uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits}
   k = zero(rate_prototype.x[1])
-  c1 = exp(-alg.gamma*dt)
-  c2 = sqrt(1 - alg.scale_noise*c1^2) # if scale_noise == false, c2 = 1
-  BAOABConstantCache(k, uEltypeNoUnits(1//2), uEltypeNoUnits(c1), uEltypeNoUnits(c2))
+  if typeof(alg.gamma) <: AbstractMatrix
+      c1 = exp(-alg.gamma*dt)
```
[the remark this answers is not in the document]
HadrienNU: There are chosen as constant for efficiency in the constant dt case, as it is the most common case. Should I pass them as function anyway? Knowing that I have as plan for the future to implement the case of gamma being a function of the current state, in which case c1 and c2 would be function of both dt and gamma. 
The alternative is to add to the documention (to be written) "Fixed timestepping only." as in the SplitODE solvers for example (https://docs.sciml.ai/DiffEqDocs/stable/solvers/split_ode_solve/)

src/caches/dynamical_caches.jl
```diff
@@ -1,42 +1,188 @@
...
+  c₂::uCoeffType
+  σ::uCoeffMType
+  tmp::uTypeCombined
+end
+
+function alg_cache(alg::ABOBA,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,jump_rate_prototype,::Type{uEltypeNoUnits},::Type{uBottomEltypeNoUnits},::Type{tTypeNoUnits},uprev,f,t,dt,::Type{Val{false}}) where {uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits}
+  k = zero(rate_prototype.x[1])
+
+  if typeof(alg.gamma) <: AbstractMatrix
+      c₂ = exp(-alg.gamma*dt)
```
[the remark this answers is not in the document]
HadrienNU: Good points, no particular reason, I will change this

test/sde/sde_dynamical.jl
HadrienNU: This comment comes from the previous implementation of BAOAB #397 as the tests are more or less the same and were copied up for ABOBA and OBABO, so I don't really know and I did not try myself, but I left the comment

src/algorithms.jl
HadrienNU: I will

src/caches/dynamical_caches.jl
```diff
@@ -1,42 +1,188 @@
...
+end
+
+function alg_cache(alg::OBABO,prob,u,ΔW,ΔZ,p,rate_prototype,noise_rate_prototype,jump_rate_prototype,::Type{uEltypeNoUnits},::Type{uBottomEltypeNoUnits},::Type{tTypeNoUnits},uprev,f,t,dt,::Type{Val{false}}) where {uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits}
+  k = zero(rate_prototype.x[1])
+  gt = zero(noise_rate_prototype)
+  half=uEltypeNoUnits(1//2)
+
+  if typeof(alg.gamma) <: AbstractMatrix
+      c₂ = exp(-alg.gamma*half*dt)
+      σ = cholesky(I - alg.scale_noise*c₂*transpose(c₂)).U
```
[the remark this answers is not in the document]
HadrienNU: I know this reference in the more general case of extended variables (but put the number of extra variable to zero and you're back into the normal Langevin case) https://doi.org/10.1021/ct900563s I can try to look for a specific reference for the Langevin case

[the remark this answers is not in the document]
HadrienNU: I have one for Langevin dynamics https://www.mdpi.com/1099-4300/19/12/647

HadrienNU:
I change the convergence tests (and separate also the tests into convergence and interface tests).

I did not update the c1 and c2 as function as @frankschae requested yet, waiting the reply to my comment

test/sde/sde_dynamical.jl
frankschae: I think we want to change it to 2, leave the allowed range as it is, and mark the test as broken (using `@test_broken`). Might be that it takes a lot of playing around with the `dts` range to see it.

src/caches/dynamical_caches.jl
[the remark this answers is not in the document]
ChrisRackauckas: Generally symplectic integrators lose the symplectic property if you adapt the time steps. Even the ODE solver ones are only fixed time step. But indeed we might as well make it safe for future work.

test/sde/sde_dynamical.jl
```diff
@@ -3,45 +3,130 @@ Random.seed!(1)
...
-    dts = (1/2) .^ (8:-1:4)
 
-    # Can't use NoiseGrid as noise is not generated with the correct size in convergence.jl. We require noise with shape of v.
-    sim1  = analyticless_test_convergence(dts,prob1,BAOAB(gamma=γ),(1/2)^10;trajectories=Int(1e2),use_noise_grid=false)
-    @test abs(sim1.𝒪est[:weak_final]-1) < 0.3
+    sol1 = solve(prob1,ABOBA(gamma=gamma_mat);dt=1/10,save_noise=true)
+    prob2 = DynamicalSDEProblem(f1_harmonic_iip,f2_harmonic_iip,g_iip,v0,u0,(0.0,5.0); noise=NoiseWrapper(sol1.W))
+    sol2 = solve(prob2,ABOBA(gamma=gamma_mat);dt=1/10)
+
+    @test sol1[:] ≈ sol2[:]
```
frankschae, line 84: oh I didn't see those tests before -- they are also iip vs oop, right? Please remove then the additional testset I pushed below. (I thought not all of them are tested because it showed that many lines were not covered by the tests in the CI coverage report.)

test/sde/sde_convergence_dynamical.jl
```diff
@@ -0,0 +1,31 @@
+using StochasticDiffEq, DiffEqNoiseProcess, Test, DiffEqDevTools, Random
```
frankschae, line 1: Can you move this file to the weak tests folder `weak_convergence`?

And also move the tests to buildkite by adding them as an additional group to:
```julia
if !is_APPVEYOR && GROUP == "WeakConvergence7"
    @time @safetestset "Dynamical SDE Tests" begin include("weak_convergence/sde_convergence_dynamical.jl") end
  end
```
https://github.com/SciML/StochasticDiffEq.jl/blob/master/test/runtests.jl

and also here: https://github.com/SciML/StochasticDiffEq.jl/blob/master/.buildkite/pipeline.yml

src/caches/dynamical_caches.jl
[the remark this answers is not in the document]
HadrienNU: On the efficiency question:  If you have a matricial gamma, if you have to do a matrix exponential and a chlolesky at each timestep you increase the computation time by a few percent.

At least on my case, the typical use case of this type of dynamics is to infers it from larger system (molecular dynamics) that are too expensive to run for long trajectories and run the inferred dynamics for very long times, so even a few percent increase of runtime could lead to extra hours of computation.

But ok, I will implement those coefficients as function, I could always add some caching behavior latter one if this becomes a performance issue. To be honest, I have never really thought to the possibility of time-stepping (losing the symplectic property is certainly an issue in my case), thanks for the reference, I will look at it.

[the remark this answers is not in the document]
frankschae: oh, then it might also be enough to check this and throw an error (if nobody would ever want to/should change it)? I think 
- fixing it and throwing an error (or documenting it very explicitly maybe) or 
-  allowing to adapt the time steps 
both sound good to me.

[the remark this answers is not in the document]
frankschae: > On the efficiency question: If you have a matricial gamma, if you have to do a matrix exponential and a chlolesky at each timestep you increase the computation time by a few percent.

I can see this point. But the cholesky/exponential also needs to be recomputed for fixed-time steps when gamma depends on the state, right? (so then it only hurts a few percent in the case of large matrix-valued gamma and fixed dt -- I have no intuition if that's a common use case.)

frankschae (approved):

frankschae:
```julia
Unexpected Pass
 Expression: abs(sim1.𝒪est[:weak_final] - 2) < 0.5
 Got correct result, please change to @test if no longer broken.
```

:-D

HadrienNU:
By the way, did you know a way to reduce the frequency of  the log output?

> [ Info: Monte Carlo iteration: 90088/100000

The logs start to be very heavy when reaching 100000 iterations

frankschae:
No but I think a PR would be very welcome. You could add a `verbose` kwarg to 

https://github.com/SciML/DiffEqDevTools.jl/blob/5a4439f33ddcc5c1ca561cad125f77e01015af72/src/convergence.jl#L121

src/caches/dynamical_caches.jl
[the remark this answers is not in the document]
HadrienNU: That make sense, I will implement those as function, will do it this week.

ChrisRackauckas:
Closing as part of StochasticDiffEq.jl archival. SDE solvers have been migrated to OrdinaryDiffEq.jl (see https://github.com/SciML/OrdinaryDiffEq.jl/pull/3154). The dynamical integrators (ABOBA, OBABO) should be re-implemented in the new OrdinaryDiffEq.jl subpackage structure if still desired.

PR #551, opened 2023-10-31
chore: bump RecursiveArrayTools compat

AayushSabharwal:


ChrisRackauckas:
It very much surprises me that the tests didn't use linear indexing at all?

PR #582, opened 2024-08-30
`alg_interpretation` from SciMLBase

prbzrg:
We need this for https://github.com/SciML/SciMLSensitivity.jl/pull/1096
And this needs https://github.com/SciML/SciMLBase.jl/pull/775 to be merged first. ✔️


ChrisRackauckas:
Actually now would be the perfect time to make that an EnumX.

prbzrg:
OK, after the tests passed. Let's do it in a new PR.

ChrisRackauckas:
This needs a bump in SciMLBase

PR #584, opened 2024-09-02
remove precs

oscardssmith:
for comaptibility with https://github.com/SciML/OrdinaryDiffEq.jl/pull/2318. Unfortunately, I don't believe there is a way to correctly sequence these to prevent temporary breakage.


oscardssmith:
I've discovered that this actually can be sequenced after the OrdinaryDiffEq side by making a slight modification to that PR.

PR #603, opened 2025-04-06
CaoTauLeaping Implementation

sivasathyaseeelan:
```julia
using StochasticDiffEq, JumpProcesses, DiffEqBase, Statistics, Plots
using Test, LinearAlgebra

function regular_rate(out,u,p,t)
    out[1] = (0.1/1000.0)*u[1]*u[2]
    out[2] = 0.01u[2]
end

const dc = zeros(3, 2)
dc[1,1] = -1
dc[2,1] = 1
dc[2,2] = -1
dc[3,2] = 1

function regular_c(du,u,p,t,counts,mark)
    mul!(du,dc,counts)
end

rj = RegularJump(regular_rate,regular_c,2)
jumps = JumpSet(rj)
iip_prob = DiscreteProblem([999.0,1,0],(0.0,250.0))
jump_iipprob = JumpProblem(iip_prob,Direct(),rj)

N = 100
sol = solve(EnsembleProblem(jump_iipprob),CaoTauLeaping();dt=1.0,trajectories = N)
plot(sol)
```

Screenshot from 2025-04-06 10-57-11

## Checklist

- [X] Appropriate tests were added
- [X] Any code changes were done in a way that does not break public API
- [X] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [X] Any new documentation only uses public API

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -847,7 +847,45 @@ SKenCarp(;chunk_size=0,autodiff=true,diff_type=Val{:central},
 
 # Jumps
 
-gillespie2001 = """@article{gillespie2001approximate,
+function TauLeaping_docstring(
+        description::String,
+        name::String;
+        references::String = "",
+        extra_keyword_description::String = "",
+        extra_keyword_default::String = "")
+    keyword_default = """
+        adaptive = true,
+        """ * "\n" * extra_keyword_default
+
+    keyword_default_description = """
+    - `adaptive`: Boolean to enable/disable adaptive step sizing. When `true`, the step size `τ` is adjusted dynamically based on error estimates or bounds. Defaults to `true`.
+    """ * "\n" * extra_keyword_description
+
+    docstring = """
+    $description
+
+    ### Algorithm Type
+    Stochastic Jump Method
+
+    ### References
+    $references
+
+    ### Keyword Arguments
+    $keyword_default_description
+
+    ### Default Values
+    $keyword_default
+    """
+    return docstring
+end
+
+@doc TauLeaping_docstring(
+    "An explicit tau-leaping method for stochastic jump processes with optional post-leap step size adaptivity. " *
+    "This algorithm approximates the stochastic simulation algorithm (SSA) by advancing the system state over " *
+    "a fixed time step `τ` using Poisson-distributed jump counts based on initial propensities. When `adaptive=true`, " *
+    "it adjusts `τ` dynamically based on post-leap error estimates derived from propensity changes.",
+    "TauLeaping",
+    references = """@article{gillespie2001approximate,
     title={Approximate accelerated stochastic simulation of chemically reacting systems},
     author={Gillespie, Daniel T},
     journal={The Journal of Chemical Physics},
@@ -855,9 +893,33 @@ gillespie2001 = """@article{gillespie2001approximate,
     number={4},
     pages={1716--1733},
     year={2001},
-    publisher={AIP Publishing}}"""
+    publisher={AIP Publishing}}""",
+    extra_keyword_description = """
+    - `dtmax`: Maximum allowed step size.
+    - `dtmin`: Minimum allowed step size.
+    - `qmax`: Maximum step size increase factor.
+    - `qmin`: Minimum step size reduction factor.
+    - `gamma`: Safety factor for step size adjustment.
+    """,
+    extra_keyword_default = """
+    dtmax = 10.0,
+    dtmin = 1e-6
+    """)
+struct TauLeaping <: StochasticDiffEqJumpAdaptiveAlgorithm
+    adaptive::Bool
+end
 
-cao2006 = """@article{cao2006efficient,
+function TauLeaping(; adaptive=true)
+    TauLeaping(adaptive)
+end
+
+@doc TauLeaping_docstring(
+    "An adaptive tau-leaping method for stochastic jump processes that selects the step size `τ` prior to each leap " *
+    "based on bounds on the expected change in state variables. Introduced by Cao et al., this method ensures stability " *
+    "and accuracy by constraining the relative change in propensities, controlled by the `epsilon` parameter. " *
+    "When `adaptive=false`, a fixed step size is used.",
+    "CaoTauLeaping",
+    references = """@article{cao2006efficient,
     title={Efficient step size selection for the tau-leaping simulation method},
     author={Cao, Yang and Gillespie, Daniel T and Petzold, Linda R},
     journal={The Journal of Chemical Physics},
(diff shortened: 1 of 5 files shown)


src/algorithms.jl
```diff
@@ -847,8 +847,16 @@ SKenCarp(;chunk_size=0,autodiff=true,diff_type=Val{:central},
...
 # Jumps
 
-struct TauLeaping <: StochasticDiffEqJumpAdaptiveAlgorithm end
-struct CaoTauLeaping <: StochasticDiffEqJumpAdaptiveAlgorithm end
+struct TauLeaping <: StochasticDiffEqJumpAdaptiveAlgorithm
+  adaptive::Bool
+end
+TauLeaping(; adaptive=true) = TauLeaping(adaptive)
+
+struct CaoTauLeaping <: StochasticDiffEqJumpAdaptiveAlgorithm
```
ChrisRackauckas, line 855: add the reference to the docstring. See OrdinaryDiffEq for how this is usually done

src/algorithms.jl
```diff
@@ -847,8 +847,102 @@ SKenCarp(;chunk_size=0,autodiff=true,diff_type=Val{:central},
...
+    publisher={AIP Publishing}}""",
+    extra_keyword_description = """
+    - `epsilon`: Tolerance parameter controlling the relative change in state variables for step size selection.
+    - `dtmax`: Maximum allowed step size.
+    - `dtmin`: Minimum allowed step size.
+    """,
+    extra_keyword_default = """
+    epsilon = 0.03,
+    dtmax = 10.0,
+    dtmin = 1e-6
```
ChrisRackauckas, line 936: these don't belong here

src/integrators/stepsize_controllers.jl
```diff
@@ -19,27 +19,85 @@ end
...
     return integrator.dt * integrator.q
+  else
+    return integrator.dt
+  end
 end
 
 function step_reject_controller!(integrator::SDEIntegrator, alg::TauLeaping)
+  if alg.adaptive
     integrator.dt = integrator.opts.gamma * integrator.dt / integrator.EEst
+  end
```
ChrisRackauckas, line 37: why is this needed?

src/integrators/stepsize_controllers.jl
```diff
@@ -19,27 +19,85 @@ end
...
+
+  # Infer ν_ij using c by applying unit counts for each reaction
+  num_reactions = length(rate)
+  ν = zeros(eltype(u), length(u), num_reactions)
+  unit_counts = zeros(eltype(rate), num_reactions)
+  for j in 1:num_reactions
+    unit_counts[j] = 1
+    c(ν[:, j], u, p, t, unit_counts, nothing)  # ν[:, j] is the change vector for reaction j
+    unit_counts[j] = 0  # Reset
+  end
```
ChrisRackauckas, line 71: This is fine for now, but we really should be storing that differently in the jumps. @isaacsas this is worth thinking about.

src/integrators/stepsize_controllers.jl
```diff
@@ -19,27 +19,85 @@ end
...
+
+  # Compute μ_i and σ_i^2
+  for i in eachindex(u)
+    for j in 1:num_reactions
+      ν_ij = ν[i, j]
+      mu[i] += ν_ij * rate[j]
```
ChrisRackauckas, line 77: The formula uses the derivative of the rate function, not the rate itself.

src/integrators/stepsize_controllers.jl
```diff
@@ -19,27 +19,85 @@ end
...
+      sigma2[i] += ν_ij^2 * rate[j]
+    end
+  end
+
+  # Compute τ per species
+  ϵ = alg.epsilon
+  τ_vals = similar(u, Float64)
+  for i in eachindex(u)
+    max_term = max(ϵ * u[i], 1.0)
```
ChrisRackauckas, line 86: This is supposed to use the sum of rates term, not u[i], also why the 1.0?

ChrisRackauckas: Oh I see.

ChrisRackauckas: this is missing the g_i factor

ChrisRackauckas:
There are 3 schemes in the paper https://people.cs.vt.edu/~ycao/publication/newstepsize.pdf, can you implement them as different algorithms (reusing most of the code, just different controller dispatches) so we can compare them?

src/integrators/stepsize_controllers.jl
isaacsas: I think it would be worthwhile to make leaping work with individually defined jumps too. Split-step methods require that anyways.

But that wouldn’t solve this issue for general non-mass action jumps.

ChrisRackauckas: and we'd need two implementations of each leaping? I'm thinking the interface should maybe require more information in the jump building, under the assumption users will be using something like Catalyst rather than building most jumps by hand. To do some of this optimally we have to shift been a few different representations, where for a user this would be redundant information to the solver it's different views that have different optimal uses.

isaacsas: For methods that work on the full vector at once, can’t we can manually build the needed function from the list of jumps? So those methods can stick with the current interface, but then we could support methods that really need jump-by-jump rate/affect access too. 

But yes, Catalyst should build the needed inputs for different representations.

ChrisRackauckas: > For methods that work on the full vector at once, can’t we can manually build the needed function from the list of jumps?

You get a lot of compile time trying to inline all of the small functions into the large one. Also there's a lack of fusion. So it's quite slow to do it that way.

I think the point is, some times you need jump-by-jump information, and sometimes you want the aggregate information (all tau leap steppers will use an aggregate function a lot of the time), and those are the same information in two very different computational representations, and building one from the other isn't easy to do from the implementation but it is trivial to do symbolically.

isaacsas: Definitely, so it should be preferred that methods that work on the full vector get a single input function, but having a fallback would be useful. Catalyst/MTK, should just generate both inputs (or whichever is most appropriate for the selected integrator -- we can add a function to query the preferred input type).

ChrisRackauckas:
Closing as part of StochasticDiffEq.jl archival. SDE solvers have been migrated to OrdinaryDiffEq.jl (see https://github.com/SciML/OrdinaryDiffEq.jl/pull/3154). The CaoTauLeaping implementation should be re-done in the new OrdinaryDiffEq.jl subpackage structure if still desired. See https://github.com/SciML/OrdinaryDiffEq.jl/issues for the migrated issue tracker.

PR #669, opened 2026-01-15
SciMLLogging Integration

jClugstor:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context

Adds a SciMLLogging verbosity specifier

### DIFF
--- a/src/initdt.jl
+++ b/src/initdt.jl
@@ -20,12 +20,12 @@ function sde_determine_initdt(
         f₀ = f(u0, p, t)
         if any(x -> any(isnan, x), f₀)
             @SciMLMessage("First function call for f produced NaNs. Exiting.",
-                integrator.opts.verbose, :function_NaN)
+                integrator.opts.verbose, :init_NaN)
         end
         g₀ = 3g(u0, p, t)
         if any(x -> any(isnan, x), g₀)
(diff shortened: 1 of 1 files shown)


jClugstor:
Problem is (this goes for DelayDiffEq as well) that it uses OrdinaryDiffEqCore functions that take the integrator, and have an `@SciMLMessage` call that expects the verbosity object in the integrator options to have the same fields as `ODEVerbosity`. 

So two options:
A) Make sure that SDEVerbosity always has the exact same toggles as ODEVerbosity, plus the SDE specific ones. 
B) Have ODEVerbosity as a field in SDEVerbosity, then forward getproperty. 

I thought option B would be more maintainable, but of course it can have some inference issues. I probably didn't see any because I didn't JET test on LTS. 

I think for now I'll switch to option A, and we'll just have to make sure that when an option is added to ODEVerbosity it also gets added to SDEVerbosity.

PR #676, opened 2026-02-18
Add callable tstops support (SymbolicTstops) for SDE/RODE solvers

isaacsas:
Enable late-binding tstops for all StochasticDiffEq algorithms, allowing callable tstops (e.g. SymbolicTstops from ModelingToolkit) that are evaluated after callback initialization when parameters are finalized.

- Add allows_late_binding_tstops trait for SDE and RODE algorithm types
- Stash callable tstops before heap init, evaluate after callback init
- Store callable in tstops_cache so reinit! can re-evaluate with potentially updated parameters
- Guard isempty(tstops) check against callable tstops
- Add comprehensive tests (45 tests covering multiple solvers, reinit!, parameter changes, callbacks, in-place/out-of-place, RODE)

This mimics the way they are handled in OrdinaryDiffEq.

### DIFF
--- a/test/callable_tstops_tests.jl
+++ b/test/callable_tstops_tests.jl
@@ -167,19 +167,26 @@ using SDEProblemLibrary: prob_sde_linear
     end
 
     @testset "Callable tstops with callbacks" begin
-        f(u, p, t) = -p[1] * u
-        g(u, p, t) = 0.1 * u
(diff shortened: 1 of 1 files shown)


isaacsas:
If tests pass I think this should be good, but let me know if you see any other issues.

PR #681, opened 2026-02-22
Add integrator-level RNG support (Phase 1, Step 3)

isaacsas:
Add `rng` kwarg to `solve`/`init`/`reinit!` for explicit RNG control, replacing the seed-only workflow. The integrator becomes the single source of truth for RNG state via a new `rng` field and the SciMLBase `has_rng`/`get_rng`/`set_rng!` interface.

Key changes:
- SDEIntegrator: add RNGType type param, `rng` and `user_provided_noise` fields
- `_resolve_rng` helper: resolves rng/seed priority with TaskLocalRNG conversion
- Wire `_rng` through all WienerProcess/CompoundPoissonProcess construction
- `set_rng!`: syncs framework noise W.rng but respects user-provided noise
- Guard noise process reseed when user provides explicit RNG
- Validate `rng` kwarg is AbstractRNG
- Bump SciMLBase compat to 2.142.1
- Add comprehensive test suite (test/rng_integrator_tests.jl)
- Add RNG Control documentation section


isaacsas:
@ChrisRackauckas I think I'm happy with this unless you have feedback.

isaacsas:
I realized that with the JumpProcesses PR https://github.com/SciML/JumpProcesses.jl/pull/557/changes the combination of these two will lead to a precedence change on the rng so I have to think through how to handle this.

PR #691, opened 2026-03-07
Fix memory allocation TODO in KenCarp step 2 and 4 by introducing chi…

Emanuele-Elias:
## Checklist

- [ ] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] The new code follows the [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context

Hi everyone! This is my first PR to SciML, aiming to prepare for GSoC.

I addressed the `TODO` left in `src/perform_step/kencarp.jl` inside `perform_step!` for `SKenCarpCache`. 
Previously, `z₄` was being used as temporary storage for `g1 * chi2` during Step 2 and Step 4, which could lead to it being overwritten and failing to properly repeat on a step fail. 

**Changes made:**
- Added a dedicated `chi2g1::uType` cache variable in the `SKenCarpCache` struct.
- Pre-allocated `chi2g1 = similar(u)` inside `alg_cache`.
- Replaced the temporary usage of `z₄` with `chi2g1` in `kencarp.jl`.
- Ran `Runic` to format the code to adhere to the SciML Style Guide.

*Note on checklist: No new tests or docs were added because this is an internal memory/cache optimization and the existing test suite already covers this algorithm. Local tests passed (the only failure was `EulerHeun sparse alloc`, which seems to be a pre-existing issue on master).*

### DIFF
--- a/src/caches/kencarp_caches.jl
+++ b/src/caches/kencarp_caches.jl
@@ -39,6 +39,7 @@ end
     chi2::randType
     g1::rateNoiseType
     g4::rateNoiseType
+    chi2g1::uType
 end
 
 u_cache(c::SKenCarpCache) = (c.z₁, c.z₂, c.z₃, c.z₄, c.nlsolver.dz)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
The todo was about g1 though? https://github.com/SciML/StochasticDiffEq.jl/pull/691/changes#diff-455cecbc916082b6b8109009404a720ac2e589ba0c966a92e4f5176e49ef4a85R309-L314

Emanuele-Elias:
I've just pushed a new commit that corrects my previous chi2g1 changes. Instead, I wrapped g(g1, ...) in an if !repeat_step && !integrator.last_stepfail block to prevent the repeated computation. To make this work, I also added a gtmp cache to replace g1 .-= g4 with @.. gtmp = g1 - g4 in Step 4 so g1 isn't destroyed with step failure. let me know if now it's ok and thank you @ChrisRackauckas for the guidance!!

ChrisRackauckas:
To fix the formatting and add an appropriate test, your changes were pulled into https://github.com/SciML/StochasticDiffEq.jl/pull/698 and merged. Thanks!

